PR #1, opened 2019-01-09
save/check GALE01r2.ini hash

hosaka-corp:
The checking probably doesn't belong in `DolphinManager.js` - probably better to do it on boot and then inform the user.


hosaka-corp:
I think verifying disc images has been solved recently, so this is probably irrelevant now.

PR #6, opened 2019-03-19
Associate the replay launcher with the .slp extension

NikhilNarayana:
Tested in windows so far.


app/components/PageWrapper.js
```diff
@@ -29,11 +30,13 @@ class PageWrapper extends Component {
   componentDidMount() {
     ipcRenderer.on('update-downloaded', this.onAppUpgrade);
     ipcRenderer.on('play-replay', this.onPlayReplay);
+    ipcRenderer.on('play-local-replay', (event, slppath) => { this.onPlayLocalReplay(slppath); });
   }
 
   componentWillUnmount() {
     ipcRenderer.removeListener('update-downloaded', this.onAppUpgrade);
     ipcRenderer.removeListener('play-replay', this.onPlayReplay);
+    ipcRenderer.removeListener('play-local-replay', (event, slppath) => { this.onPlayLocalReplay(slppath); });
```
JLaferri: Does removeListener work if you pass it a different anonymous function definition?

NikhilNarayana: I dunno. I just kinda copied how you wrote the URI handler. It never complained when I tested it.

package.json
```diff
@@ -68,13 +68,25 @@
...
+        "icon": "./resources/icon.icns"
+      },
       "category": "your.app.category.type",
       "target": [
         "zip",
         "dmg"
       ]
     },
     "win": {
+      "fileAssociations": {
```
JLaferri: I think potentially this could be at the root level to prevent needing to re-define for each OS? Wonder if you can just overwrite the `ico` and `icns` icon at the OS level

app/main.dev.js
```diff
@@ -82,8 +82,24 @@ const handleSlippiURIAsync = async (url) => {
   // Specifying a base will provide sane defaults if the input is null or wrong
   const myUrl = new URL(url, `${slippiProtocol}://null`);
   log.info(`protocol: ${myUrl.protocol}, hostname: ${myUrl.hostname}`);
-  if (myUrl.protocol !== `${slippiProtocol}:`) {
-    return;
+  if (myUrl.protocol !== `${slippiProtocol}:` && fs.existsSync(url)) {
+    const wait = ms => new Promise((resolve) => setTimeout(resolve, ms));
```
JLaferri: This looks like duplicate logic, would be nice if we could figure out a clean way to DRY it

package.json
NikhilNarayana: Made it root level, but overwriting the icon requires all the data. Luckily we only have to do that for mac I think

app/main.dev.js
```diff
@@ -75,15 +75,33 @@ const installExtensions = async () => {
   ).catch(console.log);
 };
 
+const waitForMainWindow = () => new Promise(async (resolve) => {
```
NikhilNarayana: This whole function was a doozy but it works. I had to disable linting on some lines cause I ended up in a deadlock of lint errors.

package.json
```diff
@@ -67,7 +67,17 @@
         }
       ]
     },
+    "fileAssociations": {
```
NikhilNarayana: I managed to make fileAssociations a root level key, but I couldn't rewrite the icon for mac unless I added the ext key also.

app/components/PageWrapper.js
```diff
@@ -29,11 +29,13 @@ class PageWrapper extends Component {
   componentDidMount() {
     ipcRenderer.on('update-downloaded', this.onAppUpgrade);
     ipcRenderer.on('play-replay', this.onPlayReplay);
+    ipcRenderer.on('play-local-replay', this.onPlayLocalReplay);
   }
 
   componentWillUnmount() {
     ipcRenderer.removeListener('update-downloaded', this.onAppUpgrade);
     ipcRenderer.removeListener('play-replay', this.onPlayReplay);
+    ipcRenderer.removeListener('play-local-replay', this.onPlayLocalReplay);
```
NikhilNarayana: No longer an anonymous function because I realized that the listener passes variables

app/main.dev.js
```diff
@@ -121,21 +139,7 @@ const handleSlippiURIAsync = async (url) => {
...
-    }
-
-    if (retryIdx === 100) {
-      log.warn("Timed out waiting for mainWindow to exist.");
-      return;
-    }
-
-    log.info(`Found mainWindow after ${retryIdx} tries.`);
-    mainWindow.webContents.send("play-replay");
+    waitForMainWindow().then(() => { mainWindow.webContents.send("play-replay", url); });  // eslint-disable-line
```
JLaferri: This actually changes the functionality somewhat. Previously if waiting for the main window timed out, we would not try to play the replay, with the way it's written now, the send will still occur. The timeout should probably be converted to an error inside the promise function.

package.json
```diff
@@ -67,7 +67,17 @@
...
+        "ext": "slp",
+        "name": "slp",
+        "description": "Slippi File Format",
+        "icon": "./resources/icon.ico"
```
JLaferri: could probably add `"role": "viewer"` here too (not sure what it does but we did it for URI's, it's mac only)

NikhilNarayana:
Already pulled into feature/file-assoc

PR #9, opened 2019-04-07
Add console name to game details if platform is nintendont

NikhilNarayana:


app/components/stats/GameProfile.js
```diff
@@ -192,6 +192,16 @@ export default class GameProfile extends Component {
       },
     ];
 
+    if (platform === "nintendont") {
+      const nick = 
+        _.get(this.props.store, ['game', 'metadata', 'consoleNick']);
```
JLaferri: I'd just ignore the platform and instead just check if `nick` is truthy to decide whether to push

JLaferri: This would also fix the case where this would show up blank for people that havent set the console nick yet

PR #11, opened 2019-04-10
OBS Integration

NikhilNarayana:


app/domain/SlpFileWriter.js
```diff
@@ -37,6 +46,59 @@ export default class SlpFileWriter {
...
   updateSettings(settings) {
     this.folderPath = settings.targetFolder;
+    this.obsIP = settings.obsIP;
+    this.obsSourceName = settings.obsSourceName;
+  }
+
+  async connectOBS() {
+    if (this.obsIP && this.obsSourceName) {
+      await this.obs.connect({address: this.obsIP});
+      this.scenes = await this.obs.send("GetSceneList");
```
JLaferri: `const obsScenes =` would probably be better here

NikhilNarayana: But I use that in a different function.

app/domain/SlpFileWriter.js
```diff
@@ -37,6 +46,59 @@ export default class SlpFileWriter {
...
+  }
+
+  disconnectOBS() {
+    this.obs.disconnect();
+  }
+
+  setStatus(value) {
+    this.statusOutput.status = value;
+    console.log(`Status changed: ${value}`);
+    for (let i = this.scenes.length - 1; i >= 0; i -= 1) {
```
JLaferri: what happens if scenes was never set? Also I tend to avoid for loops in javascript and tend to prefer lodash instead.

```
const scenes = this.scenes || [];
const pairs = _.flatMap(scenes, (scene) => {
  const sources = scene.sources || [];
  return _.map(sources, (source) => { scene: scene, source: source };
});
_.forEach(pairs, (pair) => {
  const res = this.obs.send(
    "SetSceneItemProperties", 
    {"scene-name": this.scenes[j].name, "item": this.obsSourceName, "visible": value}
  );
  console.log(res);
});
```

app/domain/SlpFileWriter.js
```diff
@@ -2,18 +2,27 @@ import _ from 'lodash';
...
 export default class SlpFileWriter {
   static commands = {
     CMD_RECEIVE_COMMANDS: 0x35,
+    CMD_GAME_START: 0x36,
     CMD_RECEIVE_POST_FRAME_UPDATE: 0x38,
     CMD_RECEIVE_GAME_END: 0x39,
   }
 
-  constructor(folderPath, onFileStateChange) {
+  constructor(folderPath, onFileStateChange, obsIP, obsSourceName) {
```
JLaferri: I think id prefer is the first arg was changed to an object called settings which would have fields for folderPath, obsIP, etc. This is starting to be a lot of args

app/domain/SlpFileWriter.js
```diff
@@ -37,6 +46,71 @@ export default class SlpFileWriter {
...
+        this.setStatus(false);
+      }, 100);
+    }
+
+    if (this.statusOutput.status) {
+      // If game is currently active, reset the timer
+      setTimer();
+      return;
+    }
+    if (this.currentFile.metadata.lastFrame < -70) {
```
JLaferri: maybe this check should be before the previous if block? There's no reason to set the timer if we aren't setting the status to true right?

app/components/Console.js
```diff
@@ -587,7 +581,48 @@ export default class Console extends Component {
...
+      { menuItem: "Advanced", render: () => <Tab.Pane>
+        <div className={styles['description']}>
+          <strong>Only modify if you know what you doing.</strong>&nbsp;
+          OBS integration lets you select the dolphin instance to be shown
+          if the game is active and hidden if the game is inactive.
+          You must download the &nbsp;
+          <a href="https://github.com/Palakis/obs-websocket">OBS Websocket Plugin</a>&nbsp;
+          for this feature to work.
+        </div>
+        <br/>
```
JLaferri: I usually tend to prefer using margin styles instead of `<br/>` elements but I also don't mind that much

app/domain/SlpFileWriter.js
NikhilNarayana: agreed, I moved them around

app/components/Console.js
NikhilNarayana: I figured out how to add a spacer for this section and then tab area as a whole so we should be good there.

PR #19, opened 2019-04-27
Update OBS Autoswitch

NikhilNarayana:
* Rewrite how we store the pairs so we spend less time looking for the sources in the array
* Update the pairs at the end of the game so users don't have to disconnect and reconnect to auto switch new SceneItems (this is a stopgap until Palakis releases the next version of OBS Websocket that supports better events)
* Update the timer at game end to 700ms and the frame we start on to -60

### DIFF
--- a/app/domain/SlpFileWriter.js
+++ b/app/domain/SlpFileWriter.js
@@ -51,12 +51,23 @@ export default class SlpFileWriter {
     this.obsPassword = settings.obsPassword;
   }
 
+  getSceneSources = async (data) => {
+    console.log(data);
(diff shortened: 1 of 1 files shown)


app/domain/SlpFileWriter.js
```diff
@@ -51,12 +51,23 @@ export default class SlpFileWriter {
     this.obsPassword = settings.obsPassword;
   }
 
+  getSceneSources = async (data) => {
+    console.log(data);
```
JLaferri: should probably get rid of these loggers?

PR #20, opened 2019-04-29
Support the new handshake protocol

NikhilNarayana:
Initial work is done, just need to clean up the code and test a bit more


NikhilNarayana:
As of this post we have finished the actual handshake, but we need to start considering some aspects of the transfer that we didn't before.

Right now when we disconnect and reconnect during a game dolphin fails to pick up the new file and stays frozen until the next game. My current theory is that if we can somehow maintain the last file name somewhere we can append to the slp file without an issue. But after that we lose the frame count so the metadata might be wrong in the end. meta suggested we instead build the metadata on the Wii and send that over at the end of the game.

NikhilNarayana:
Merged into updated-comms branch

PR #23, opened 2019-05-25
Add support for showOnFrame and hideOnFrame

NikhilNarayana:
Advanced settings because the default config won't always work for everyone

image

image


JLaferri:
Is the screenshot of the default state active? I think it might be nice if it was populated with the default values instead and the placeholder text was more of a description under the label?

NikhilNarayana:
> Is the screenshot of the default state active? I think it might be nice if it was populated with the default values instead and the placeholder text was more of a description under the label?

First image shows the default values for the new fields and the second image shows placeholder text. If you think we need description text, I'll add it in tonight.

NikhilNarayana:
splitting this into two separate PRs

PR #27, opened 2019-06-29
Pagination of replay files

VilleMiekkoja:
Displays 30 replays per page. Easily changeable

kuva

### DIFF
--- a/app/components/FileLoader.js
+++ b/app/components/FileLoader.js
@@ -11,6 +11,7 @@ import {
   Segment,
   Message,
   Loader,
+  Pagination,
 } from 'semantic-ui-react';
 import styles from './FileLoader.scss';
 import FileRow from './FileRow';
@@ -20,6 +21,8 @@ import FolderBrowser from './common/FolderBrowser';
 import PageWrapper from './PageWrapper';
 import Scroller from './common/Scroller';
 
+const MAX_NUM_OF_FILES_PER_PAGE = 30.0
+
 export default class FileLoader extends Component {
   static propTypes = {
     // fileLoader actions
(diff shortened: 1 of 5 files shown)


app/components/FileLoader.js
```diff
@@ -282,8 +286,17 @@ export default class FileLoader extends Component {
...
+
+    const howManyLeft = (files.length - 1) - fileBeginIndex
+    let lastFileIndex
+    if (howManyLeft > MAX_NUM_OF_FILES_PER_PAGE) {
+      lastFileIndex = fileBeginIndex + MAX_NUM_OF_FILES_PER_PAGE
+    } else {
+      lastFileIndex = fileBeginIndex + howManyLeft
+    }
+
+    const rows = files.slice(fileBeginIndex, lastFileIndex).map(
```
JLaferri: What would happen in the case where you're on say, page 5 in a folder, and then you switch to a different folder that has only one page?

VilleMiekkoja: Need to check. But my guess is, that the fileloader is then completely reloaded, and so these files and pages are calculated again. But yeah need to check, good point

app/components/FileLoader.js
```diff
@@ -282,8 +286,17 @@ export default class FileLoader extends Component {
...
-    const rows = files.map(
+    const fileBeginIndex = (this.props.store.currentPage - 1) * MAX_NUM_OF_FILES_PER_PAGE
+
+    const howManyLeft = (files.length) - fileBeginIndex
+    let lastFileIndex
+    if (howManyLeft > MAX_NUM_OF_FILES_PER_PAGE) {
+      lastFileIndex = fileBeginIndex + MAX_NUM_OF_FILES_PER_PAGE
+    } else {
+      lastFileIndex = fileBeginIndex + howManyLeft
+    }
```
fztgreen, line 297: Probably should just use a ternary here, more concise
```
lastfileIndex =  fileBeginIndex + ((files.length) - fileBeginIndex) > MAX_NUM_OF_FILES_PER_PAGE ? + MAX_NUM_OF_FILES_PER_PAGE :  howManyLeft
```

JLaferri:
Closing this in favor of #58

PR #31, opened 2019-07-08
Add ability to select multiple files to queue for dolphin playback

jmlee337:
this lets you playback a whole set at once, etc.

also paves the way to do other batch operations if wanted (delete, star, relabel, etc.)

queue

### DIFF
--- a/app/components/FileLoader.js
+++ b/app/components/FileLoader.js
@@ -88,19 +88,19 @@ export default class FileLoader extends Component {
     this.props.dismissError('fileLoader-global');
   }
 
-  onSelect = (key) => {
+  onSelect = (selectedFile) => {
     const newSelections = [];
 
     let wasSeen = false;
-    this.state.selections.forEach(selection => {
-      if (selection === key) {
+    this.state.selections.forEach(file => {
+      if (file === selectedFile) {
         wasSeen = true;
         return;
       }
-      newSelections.push(selection);
+      newSelections.push(file);
     });
     if (!wasSeen) {
-      newSelections.push(key);
+      newSelections.push(selectedFile);
     }
 
     this.setState({
@@ -343,7 +343,7 @@ export default class FileLoader extends Component {
           playFile={this.props.playFile}
           gameProfileLoad={this.props.gameProfileLoad}
           onSelect={this.onSelect}
-          selectedOrdinal={this.state.selections.indexOf(file.fullPath) + 1}
+          selectedOrdinal={this.state.selections.indexOf(file) + 1}
         />
       ),
       this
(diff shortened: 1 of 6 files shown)


app/domain/DolphinManager.js
```diff
@@ -77,12 +68,38 @@ export default class DolphinManager {
...
     }
   }
 
+  async queueFiles(filePaths) {
+    const jsonString = JSON.stringify({
+      mode: "queue",
+      replay: "",
+      isRealTimeMode: false,
+      queue: filePaths.map(filePath => ({
+        path: filePath,
```
NikhilNarayana: Might be good to add these extra fields for the users that might want it

https://github.com/project-slippi/project-slippi/blob/8176c70febb46c31c914481e275e5396eb571a19/scripts/extractCombos.js#L118-L119

jmlee337: you got it

app/components/FileLoader.js
```diff
@@ -41,6 +42,24 @@ export default class FileLoader extends Component {
...
+  }
+
+  static getDerivedStateFromProps(newProps, oldState) {
+    if (newProps.store.files !== oldState.files) {
+      return {
+        files: newProps.store.files,
+        selections: [],
+      };
+    }
+    return oldState;
```
JLaferri: According to the docs you can return `null` to not change state. Also in general I don't love managing state like this. I think putting selections in redux might actually be better/more flexible. But for now this is fine.

app/components/FileLoader.js
```diff
@@ -41,6 +42,24 @@ export default class FileLoader extends Component {
     topNotifOffset: PropTypes.number.isRequired,
   };
 
+  constructor(props) {
+    super(props);
+    this.state = {
+      files: [],
+      selections: [],
+    };
+  }
```
JLaferri: Don't think you need to declare a constructor like this, I think you can just define this directly in the class:
```
state = {
 files: [],
 selections: [],
}
```

Though maybe this has limitations that I'm not aware of.

app/components/FileLoader.js
```diff
@@ -290,6 +342,8 @@ export default class FileLoader extends Component {
           file={file}
           playFile={this.props.playFile}
           gameProfileLoad={this.props.gameProfileLoad}
+          onSelect={this.onSelect}
+          selectedOrdinal={this.state.selections.indexOf(file) + 1}
```
JLaferri: This is an unnecessary O(n) operation in a loop, so worst case run time of this would be O(nm). It could be avoided by using a keyed object for selections instead where the key is the full path of the file. That would also make the "add selection" logic better as well.

app/components/FileLoader.js
jmlee337: as far as I know, it should probably be changed to that yeah

PR #35, opened 2019-08-21
feature/file-renaming

ethangamma24:
Something in the code is causing the modal to stay open after the submit button is clicked until it is clicked again, or the user clicks outside of the modal. I _believe_ it has something to do with the onSelect() function, but I'm not entirely sure. Additionally, when the modal is open, the user's clicks are still registered by the DOM in the background, so onSelect() keeps getting triggered even though the background isn't in focus, and I don't know how to fix that.


ethangamma24:
Not sure why this got closed

NikhilNarayana:
I think you picked the wrong branch, closing until the correct branch is selected.

Issue #41, opened 2019-11-16
Update node-sass dependency to 4.12.0 or 4.13.0


N0NamedGuy:
I have node 12 installed in my machine, and I am not able to build. This happens because when using node 12 the it is not possible to install/build node-sass 4.10.0.

node-sass needs to be upgrade to at least 4.12.x Keep in mind node 12 is the latest LTS version.

PR #42, opened 2019-11-16
node-sass update

N0NamedGuy:
Se #41


JLaferri:
I believe I've been using 10.15 LTS but this should be fine regardless. If I have issues might revert.

PR #44, opened 2020-01-31
Fixed issue #38, added grab release

### LINKED ISSUE #38: Kills that don't involve connected moves break the kills table
In the file below, the falco dies off of a grab release on his third stock. The punish stat for this interaction lacks any moves so the kills table fails to generate and breaks the entire stats page.

[Game_20191008T211638.zip](https://github.com/project-slippi/slippi-desktop-app/files/3704346/Game_20191008T211638.zip)

seant4:
fixes #38

### DIFF
--- a/app/components/stats/KillsTable.js
+++ b/app/components/stats/KillsTable.js
@@ -71,11 +71,17 @@ export default class KillsTable extends Component {
     const killingPunishesByEndFrame = _.keyBy(killingPunishes, 'endFrame');
     const punishThatEndedStock = killingPunishesByEndFrame[stock.endFrame];
 
-    if (!punishThatEndedStock) {
+    if(!punishThatEndedStock ) {
       return <span className={styles['secondary-text']}>Self Destruct</span>;
     }
+    
 
     const lastMove = _.last(punishThatEndedStock.moves);
+
+    if((_.last(punishThatEndedStock.moves)) == undefined){
+      return "Grab Release"
+    }
+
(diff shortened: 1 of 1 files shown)


app/components/stats/KillsTable.js
```diff
@@ -71,11 +71,17 @@ export default class KillsTable extends Component {
...
 
-    if (!punishThatEndedStock) {
+    if(!punishThatEndedStock ) {
       return <span className={styles['secondary-text']}>Self Destruct</span>;
     }
+    
 
     const lastMove = _.last(punishThatEndedStock.moves);
+
+    if((_.last(punishThatEndedStock.moves)) == undefined){
```
JLaferri: You don't need to rerun the same code that just ran. Additionally it's not necessary to check explicitly for undefined. Also never use `==`, only `===`. But in this case you should just do `if (!lastMove)`

app/components/stats/KillsTable.js
```diff
@@ -74,9 +74,14 @@ export default class KillsTable extends Component {
     if (!punishThatEndedStock) {
       return <span className={styles['secondary-text']}>Self Destruct</span>;
     }
+    
 
     const lastMove = _.last(punishThatEndedStock.moves);
-    return moveUtils.getMoveName(lastMove.moveId);
+    if (!(lastMove)) {
+      return "Grab Release"
+    }else{
```
JLaferri: You don't need this else.

PR #49, opened 2020-03-09
Add play all function

vmaggioli:
Essentially kept the "Play All" button always displaying and it enqueues all files (post filtering) for replay. Would like to know if there are any preferences on playback order, as I have to reverse the file array in order to play from oldest to newest as specified by the [trello](https://trello.com/c/Fxes0Izs/118-play-all-button-to-play-all-files-in-a-folder-in-a-row). I can see this having performance issues on a large number of replays and not really giving us significant functionality in return.

image

### DIFF
--- a/app/components/FileLoader.js
+++ b/app/components/FileLoader.js
@@ -189,6 +189,11 @@ export default class FileLoader extends Component {
     });
   }
 
+  queueAllFiles = () => {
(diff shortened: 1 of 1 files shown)


vmaggioli:
Selectable replay functionality exists now in dev, closing to switch requirements from a "play all" to a "select all". Will create a new PR after adjustments are made

PR #50, opened 2020-03-10
Add select all functionality

vmaggioli:
Originally intended to be an auto "play all" feature, but with the queue feature already being merged in I've piggybacked off of it and just made a "select all" for it.

[Trello](https://trello.com/c/Fxes0Izs/118-play-all-button-to-play-all-files-in-a-folder-in-a-row)

Pre-select
image

Post-select
image

### DIFF
--- a/app/components/FileLoader.js
+++ b/app/components/FileLoader.js
@@ -137,6 +137,11 @@ export default class FileLoader extends Component {
     });
   }
 
+  queueAllFiles = () => {
+    const filteredFiles = this.processFiles(this.state.files)
+    this.props.queueFiles(_.reverse(filteredFiles));
+  }
+
   renderGlobalError() {
     const errors = this.props.errors || {};
     const errorKey = 'fileLoader-global';
@@ -317,10 +322,14 @@ export default class FileLoader extends Component {
   }
 
   renderQueueButtons() {
-    if (this.state.selections.length === 0) {
-      return;
-    }
-    return (
+    return this.state.selections.length === 0 ? (
+      <div className={styles['queue-buttons']}>
+        <Button onClick={this.queueAllFiles}>
+          <Icon name="play circle" />
+          Play all
+        </Button>
+      </div>
+    ) : (
       <div className={styles['queue-buttons']}>
         <Button onClick={this.queueFiles}>
           <Icon name="play circle" />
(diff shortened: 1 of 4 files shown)


JLaferri:
Instead of being a constantly floating element, I kind of figured the UI should be closer to what you see in gmail.

In other words:
1. A checkbox in the top-left that resembles the other checkboxes that when clicked, will select everything
2. When something is selected, clicking that checkbox will de-select everything

vmaggioli:
Got it, I'll switch the select all button to another check box in the empty grid item next to "Details"

vmaggioli:
Force pushes were for rebase and rebase issue correction

vmaggioli:
Switched persistent button to a checkbox in the table, similar to gmail. Currently still working on select+click functionality as requested by @JLaferri.
image

image

JLaferri:
Looks good. Is de-select working? If so can probably remove the clear button.

As far as minor UI stuff, I might suggest increasing the font size a small amount of "Details" and "Time" headers and bottom aligning them

vmaggioli:
De-select is working as well, but I’d imagine we’d want to keep the clear as somebody can mark a lot of files but not all of them and then want to clear their selections

vmaggioli:
Also looking through the CSS it looks like the padding from the bottom for the headers is dynamically calculated by the table component we're using, so I'm not sure if we want to start messing with that

vmaggioli:
Adjusted table header font size from original 1 em to 1.2 em
image

JLaferri:
For the bottom align thing, you might want to browse the semantic UI docs:
https://react.semantic-ui.com/collections/table/

If you click on Table.Cell, you will see the following:
image

That means that on the HeaderCell react element you can set the property `verticalAlign="bottom"` and it will automatically use bottom alignment without writing any custom CSS. It might help to look through the docs generally because semantic UI is used fairly heavily throughout and it helps to know where you can avoid writing CSS.

vmaggioli:
Force pushed due to forgetting to remove commented out testing code

vmaggioli:
Let me know if I'm reading it wrong, but it looks like `verticalAlign` is only for `Table.Cell` and not `Table.HeaderCell`. I tried their coding examples on their and it worked fine for Cells but didn't seem to do anything for HeaderCells. Also I've added the shift select functionality, please check out the branch and tinker with it to see if it behaves how you're expecting.

JLaferri:
I tested it myself, it works on HeaderCell too. I imagine HeaderCell supports all of the Cell properties

JLaferri:
Though it does seem like `verticalAlign` bottom might be the default anyway...

PR #51, opened 2020-03-13
Stats page nav

vmaggioli:
Added buttons for navigating to previous and next matches in replay folder without leaving the stats page. Lifted filtering short/error files from component to redux store level due to have access to file list in GameProfile component.

[Trello](https://trello.com/c/UiHlb8iP/147-navigate-between-games-on-stats-page)

image

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -30,6 +30,7 @@
     "no-mixed-operators": "off",
     "no-use-before-define": "off",
     "no-multi-assign": "off",
+    "no-plusplus": "off",
     "object-shorthand": ["error", "never"],
     "one-var": "off",
     "one-var-declaration-per-line": "off",
--- a/app/actions/fileLoader.js
+++ b/app/actions/fileLoader.js
@@ -1,15 +1,18 @@
 import _ from 'lodash';
 import fs from 'fs';
 import path from 'path';
+import moment from 'moment';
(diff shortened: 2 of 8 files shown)


nachoverdon:
It'd be great if the buttons where a little bit higher up, so that it has some margin at the bottom, don't you think?

vmaggioli:
Agreed, I've been having issues trying to get the buttons to respect the same padding as the match details but for some reason it's just adding the padding onto the bottom of the buttons versus actually shifting the buttons up. Will post a follow up screenshot when I can get that working

vmaggioli:
Post adjustments
image

JLaferri:
UI looks pretty good. I think the bottom margin should match the bottom margin of the metadata display, still looks like it's less

vmaggioli:
It is less, I currently have the div containing the buttons vertically centered with the metadata div. The issue is that the buttons themselves are taller than the metadata div, so I figured distributing that by centering would like the best. I can try to get the bottoms aligned and show what it looks like

vmaggioli:
Here's bottom aligned. Don't mind the weird indent on the left, I'll fix that if we go forward with this layout.

image

vmaggioli:
New layout with bottom align and margins fixed
image

PR #52, opened 2020-03-21
Reset Dolphin Button

NikhilNarayana:
Also cleans up the code that adds the game path to dolphin by moving the bit of code to dolphin manager.

image

[Video of reset in action](https://streamable.com/4nnbe)

### DIFF
--- a/app/components/Settings.js
+++ b/app/components/Settings.js
@@ -121,7 +121,7 @@ export default class Settings extends Component {
         <LabelDescription
           label="Reset Playback Dolphin"
           description="
-            If playback of files fails to work, this button will reset the Playback
+            If replay playback fails to work, this button will reset the Playback
             Dolphin to its original state. If this doesn't fix your issue please head
             over to our Discord to receive further support
           "
--- a/app/components/Settings.js
+++ b/app/components/Settings.js
@@ -147,6 +147,7 @@ export default class Settings extends Component {
           inverted={true}
           onClick={this.showConfirmReset}
           loading={store.isResetting}
+          disabled={store.isResetting}
         />
         <Confirm
           className={styles['confirm']}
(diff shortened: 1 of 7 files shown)


app/components/Settings.js
```diff
@@ -109,11 +114,67 @@ export default class Settings extends Component {
...
+          onClick={this.showConfirmReset}
+          loading={store.isResetting}
+        />
+        <Confirm
+          className={styles['confirm']}
+          open={!!confirmShow}
+          confirmButton="Yes"
+          cancelButton="No"
+          header="Reset Dolphin?"
+          content="Are you sure you would like to reset Dolphin? This will update all the settings to the defaults"
```
JLaferri: "This will return" instead of update?

app/reducers/settings.js
```diff
@@ -119,3 +125,17 @@ function isoValidationComplete(state, action) {
     isoValidationState: action.payload.isValid,
   };
 }
+
+function toggleConfirmDialog(state, action) {
```
JLaferri: These seem more like sets than toggles. Toggle to me kind of implies the current state is switched.

app/domain/DolphinManager.js
```diff
@@ -51,6 +53,50 @@ export default class DolphinManager {
...
+    try {
+      const iniPath = path.join(dolphinPath, "User", "Config", "Dolphin.ini");
+      const dolphinINI = ini.parse(fs.readFileSync(iniPath, 'utf-8'));
+      dolphinINI.General.ISOPath0 = fileDir;
+      const numPaths = dolphinINI.General.ISOPaths;
+      dolphinINI.General.ISOPaths = numPaths !== "0" ? numPaths : "1";
+      const newINI = ini.encode(dolphinINI);
+      fs.writeFileSync(iniPath, newINI);
+    } catch (err) {
+      log.warn(`Failed to update the dolphin paths\n${err}`)
```
JLaferri: should probably throw here to not swallow error

PR #57, opened 2020-05-27
domain: handle openingBytes anywhere in buffer

NikhilNarayana:
change openingBytes => typeBytes by chopping off the first 3 values

check if the typeBytes are in order anywhere within the data buffer by converting to string and finding the substring.

Tested with console and altf4's slippicomm branch, both complete the handshake fine.


NikhilNarayana:
closing because we are just gonna force type to be the start on Dolphin's side

PR #58, opened 2020-06-13
Performance: Lazy Load Game Rows in Replay Browser

### LINKED ISSUE #53: Performance: Replay Browser's loading time is proportinal to number of replays
Currently, there's no support for the pagination/segmentation of replays. The result of this is that every replay file must be loaded before the UI becomes usable.

After using slippi for a few months (accumulating about 2500 replays), my load time is about 15-20 seconds. Adding proper pagination support would allow for a constant load time of the desktop app.

boulajp:
Fixes #53

The approach taken here is similar to an infinite scroll - we'll lazy-load game rows in chunks of a pre-configured size, whenever we've reached the end of the table that we've already loaded. We use the `Visibility` element to determine whether the element at the end of the table is visible, and then use the callback that is triggered to render another chunk of games.

### DIFF
--- a/app/components/FileLoader.js
+++ b/app/components/FileLoader.js
@@ -20,6 +20,8 @@ import FolderBrowser from './common/FolderBrowser';
 import PageWrapper from './PageWrapper';
 import Scroller from './common/Scroller';
 
+const GAME_BATCH_SIZE = 10;
+
 export default class FileLoader extends Component {
   static propTypes = {
     // fileLoader actions
@@ -319,17 +321,15 @@ export default class FileLoader extends Component {
     );
 
     const bufferMoreFiles = () => {
-      const fileBufferSize = 10;
-
       const start = filesOffset;
-      const end = Math.min(start + fileBufferSize, allFiles.length);
+      const end = Math.min(start + GAME_BATCH_SIZE, allFiles.length);
 
       if (start < allFiles.length)
       {
         const nextFilesToRender = allFiles.slice(start, end);
         this.setState({
           filesToRender: filesToRender.concat(nextFilesToRender),
-          filesOffset: end+1,
+          filesOffset: end,
         });
       }
     }
--- a/app/components/FileLoader.js
+++ b/app/components/FileLoader.js
@@ -10,6 +10,7 @@ import {
   Segment,
   Message,
   Loader,
+  Visibility,
 } from 'semantic-ui-react';
 import styles from './FileLoader.scss';
 import FileRow from './FileRow';
@@ -27,6 +28,7 @@ export default class FileLoader extends Component {
     playFile: PropTypes.func.isRequired,
     queueFiles: PropTypes.func.isRequired,
     storeScrollPosition: PropTypes.func.isRequired,
+    storeFileLoadState: PropTypes.func.isRequired,
     setStatsGamePage: PropTypes.func.isRequired,
 
     // error actions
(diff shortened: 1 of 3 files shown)


app/components/FileLoader.js
```diff
@@ -302,6 +318,28 @@ export default class FileLoader extends Component {
       this
     );
 
+    const bufferMoreFiles = () => {
+      const fileBufferSize = 10;
```
boulajp, line 322: Suggestion: Pull this out to the top of the file as a const, and give a better name like `GAME_BATCH_SIZE`

JLaferri:
Noticed you closed this. Sorry for the lack of communication, I've been quite busy.  Did this help a lot with performance for you? I'll try to test it next week and try to get it merged if you're confident in it.

boulajp:
No worries at all - I temporarily closed this PR until I understand the open-source policies that I'm subject to through my employer. I'll re-open the PR once I get an official thumbs up, which will hopefully be early next week.

boulajp:
Reopening this now.

NikhilNarayana:
Do you have any comparisons of load time with this new method?

boulajp:
Yup, I have just over 2500 replays - my load time has gone from ~23 seconds to 1 seconds

app/components/FileLoader.js
```diff
@@ -264,14 +277,17 @@ export default class FileLoader extends Component {
 
   renderFileSelection() {
     const store = this.props.store || {};
-    const files = store.files || [];
```
a-sink-a-wait, line 267: I don't think this is referenced anymore?

a-sink-a-wait:
Do we have unit tests around these `.js` files?

app/components/FileLoader.js
boulajp: What isn't referenced? allFiles is used a few lines below

a-sink-a-wait: Might have been looking at wrong commit. Still saw `const files = store.files || [];`

boulajp:
> 
> 
> Are there unit tests around these `.js` files?

I don't believe there are tests in this project.

a-sink-a-wait:
> I don't believe there are tests in this project.

Is that something worth pursuing? From my little understanding, this looks ok to me.

boulajp:
@a-sink-a-wait I'm not too familiar with front-end javascript, so I'm not too sure what it'd take to add testing to this project - though I'd imagine that there's established test patterns for React/Redux projects.

To answer your question of "whether itd be worth pursuing", I think it would be useful - my first commit actually introduced two bugs:
1. Since `Array.prototype.slice` is non-inclusive for the final index, we were skipping over the final file in a batch chunk
2. Upon changing folders, we never cleared the `fileLoadState`, and this we'd be looking at an old offset in our new folder.

I had no way to know this was an issue at check-in time, and just had a "hunch" that I should verify that the offsets were correct a few days later. Had there been a suite of tests, we could have been able to check whether `filesToRender` contained the expected files, which would have prevented the first bug.

The issue with adding tests at this point is that the project is already in a decently mature state, and it may not have been written in a very testable way (this is just my assumption as I don't know much about testing patterns in react / JS). 

@NikhilNarayana or @JLaferri could answer that question better than me.

jwsharpe:
@boulajp 
PR looks amazing, a few things:

It might be better practice to put the Visibility element as the `thead` around the file rows. This will give us better resources to hammer out some of the load 'stuttering'.
Increasing the batch size higher might give it less of a stutter.
I threw a PR on your PR for some ux changes. 
https://github.com/boulajp/slippi-desktop-app/pull/1/

boulajp:
Thanks @jwsharpe. Overall LGTM, and the UI loading you added in when I tested locally. I left a few comments that we could probably fix up before merging in. 

Also, yup I just joined the discord.

boulajp:
Any word when we could get this merged in? We've had a few people test this locally and it seems to be fine.

NikhilNarayana:
Code LGTM

Tests:
15k replays
* Loads in 1:55.43 on public
* Loads in 6.92 on branch with very minimal stuttering when scrolling fast!

Love the changes

PR #59, opened 2020-06-28
Remove checking for PUSH action on file loader menu

walz0:
From testing I found that this conditional is only preventing the user's files from loading on first use (selecting 'Select Folder' from the replay browser without an slp file path set). If the user is redirected to the settings menu without an slp file path set and then hits 'Back' after setting a file path, the files will not load because of this conditional. Functionality for storing scroll position after the user views match statistics or goes back to the main menu remain the same.


JLaferri:
If I remove this, I can't load my root folder at all.

PR #60, opened 2020-06-28
Display netplay codes on the app.

pnivlek:
Currently, this will display the netplay name and code on the game profile page. On the file list, it displays the code and falls back to the old player name logic if there is no code.

Screenshots (second image using Player 1 and 2 since this is before the update to pass nicknames for players on Dolphin):
image

image


pnivlek:
Force push to make the author of the last commit not "= <=>".

JLaferri:
lgtm, thanks!

PR #62, opened 2020-07-02
Bugfix: Subdirectories should now work correctly

### LINKED ISSUE #61: Subdirectories do not work, and the root directory's list of replays are shown no matter directory is picked.
As of the latest pull request (#58 ), sub directories do not filter anything. The store's list of files updates, but filesToRender does not.

boulajp:
Fixes #61 
Fixes subdirectory navigation after changes in #58.

These changes break save scroll positions - i.e. when we open a game, then navigate back to the replay browser, we will no longer return to the position we were on in the list of replays. This is due to the fact that now, we trigger an action (`STORE_FILE_LOAD_STATE`) upon the first render, and thus `componentDidMount` (which contains the logic to move to a scroll position) occurs before we've rendered the replays

### DIFF
--- a/app/components/FileLoader.js
+++ b/app/components/FileLoader.js
@@ -46,14 +46,8 @@ export default class FileLoader extends Component {
   constructor(props) {
     super(props);
 
-    const filesToRender = _.get(this.props.store, ['fileLoadState', 'filesToRender']) || [];
-    const filesOffset = _.get(this.props.store, ['fileLoadState', 'filesOffset']) || 0;
-
     this.state = {
       selections: [],
-      filesToRender: filesToRender,
-      filesOffset: filesOffset,
-      firstLoad: true,
     };
   }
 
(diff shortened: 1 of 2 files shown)


JLaferri:
Thanks a ton for fixing the issue as it was reported!

Prior to these changes the position restoration was working?

I know there was another active PR that removed something I had commented with regards to scroll position. Since we're looking into scroll position restoration, is there any chance we could look into this PR too? https://github.com/project-slippi/slippi-desktop-app/pull/59/files

boulajp:
Yup, scroll position was working both prior and after #58 - this PR will break it if we merge, but will fix subdirectories (it was my bad - I didn't actually know that subdirectories were supported in the replay app!). 

I'm not sure what relation that PR will have to this, but those changes, unfortunately, won't fix the issue I'll introduce, as `componentDidMount` still won't run after the rows are rendered.

JLaferri:
I'll just merge this and look into fixing the scroll position

boulajp:
Thanks! I have an idea how we can fix scroll position. I’ll look into that next week, but have opened issue #65 to track this.

PR #68, opened 2020-07-06
Spectator changes + fix performance bug

flux627:
Forked from @altf4's `spectator` branch [here](https://github.com/altf4/slippi-desktop-app/tree/spectator) which should be used with [this corresponding Ishiiruka PR](https://github.com/project-slippi/Ishiiruka/pull/98).

Includes all of his changes, along with:

- Refactor `SlpFileWriter#currentFile` to instance of new `SlpFile` class. (Open to better names for this new class.)
- Only merge `previousBuffer` if it exists
- Properly check if clients array has items with `.length`
- Only concat buffers when `fullBuffer` is accessed via getter
- Linting changes / fixes

This PR defers merging of buffer data to `fullBuffer` until it is accessed, which resolves performance issues *as long as no clients are connected via relay mode*. This problem still exists in relay mode, but will be addressed in a subsequent PR.

### DIFF
--- a/app/domain/ConsoleConnection.js
+++ b/app/domain/ConsoleConnection.js
@@ -147,9 +147,8 @@ export default class ConsoleConnection {
     // }
 
     this.connectToSpectate();
-    // TODO Turn these back on. But it was causing issues
-    // this.connectOnPort(Ports.WII_DEFAULT);
-    // this.connectOnPort(Ports.WII_LEGACY);
+    this.connectOnPort(Ports.WII_DEFAULT);
+    this.connectOnPort(Ports.WII_LEGACY);
   }
 
   connectToSpectate() {
--- a/app/domain/ConsoleConnection.js
+++ b/app/domain/ConsoleConnection.js
@@ -18,6 +18,7 @@ Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
 
 */
 
+/* eslint no-unused-vars: ["error", { "argsIgnorePattern": "^_" }] */
 
 import net from 'net';
 import inject from 'reconnect-core';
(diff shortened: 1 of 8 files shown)


altf4:
Commit a774606 breaks the functionality. I disabled the `connectOnPort()` functions because they didn't play nicely with the changes I was making.

I think what happens is that the ENet port connects just fine, but then the program tries to also connect on the two TCP ports. These fail, and it changes the connection status, breaking the existing connection. I'm not sure what the best way to handle this is? The behavior we want I guess is to try all three connection ports and accept the first one that works?

app/domain/ConsoleConnection.js
```diff
@@ -135,18 +138,107 @@ export default class ConsoleConnection {
...
     this.dolphinManager.updateSettings(connectionSettings);
 
-    if (this.port && this.port !== Ports.WII_LEGACY && this.port !== Ports.WII_DEFAULT) {
-      // If port is manually set, use that port. Don't do this if the port is set to legacy as
-      // somebody might have accidentally set it to that and they would encounter issues with
-      // the new Nintendont
-      this.connectOnPort(this.port);
-      return;
-    }
+    // if (this.port && this.port !== Ports.WII_LEGACY && this.port !== Ports.WII_DEFAULT) {
```
a-sink-a-wait, line 141: Commented code?

altf4: I meant these two lines:

https://github.com/project-slippi/slippi-desktop-app/blob/d7db6b69bcc875a20ba93ba0e019c5975995c49b/app/domain/ConsoleConnection.js#L150

flux627:
This PR still has problems, but I probably won't have time to work on it more until this weekend. If anyone else wants to carry the torch forward before then, please do so by all means.

altf4:
When trying to resume in the middle of a match (>4 mins long) I get this error:

image

Which apparently is caused by this?

https://stackoverflow.com/questions/27769842/write-after-end-error-in-node-js-webserver

NikhilNarayana:
closing in favor of the https://github.com/project-slippi/slippi-desktop-app/tree/feature/broadcast branch

PR #75, opened 2020-07-09
use file creation date in "Details" of Replay Browser if the date is not available in the metadata

### LINKED ISSUE #73: Add support for inferring time based on the filename
The number of replays with `Unknown` time has been increasing recently and it might be worth just inferring the time from the file name if the metadata is empty since most users are probably not going to rename the files. If the time can't be inferred then it will stay as `Unknown`.

Time format for replay names is `%Y%m%dT%H%M%S` on Dolphin.
For Nintendont, check this: https://github.com/project-slippi/Nintendont/blob/slippi/kernel/SlippiFileWriter.c#L83-L87 (we write the MAC address in addition to the timestamp)

davidtranhq:
Fixes #73 

This uses the `fs` module to get the creation date of the file if the metadata could not provide a start time.

### DIFF
--- a/app/components/FileRow.js
+++ b/app/components/FileRow.js
@@ -12,6 +12,7 @@ import PlayerChiclet from './common/PlayerChiclet';
 import * as timeUtils from '../utils/time';
 
 const path = require('path');
+const fs = require('fs');
 
 export default class FileRow extends Component {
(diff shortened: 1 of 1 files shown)


NikhilNarayana:
Closing in favor of #77

PR #77, opened 2020-07-09
Unknown time fix

### LINKED ISSUE #73: Add support for inferring time based on the filename

Octopirate1:
Fixes #73 

Also adds support for inferring from filename in case the file was copied or downloaded from another machine, which would affect ctime

### DIFF
--- a/app/components/FileRow.js
+++ b/app/components/FileRow.js
@@ -12,7 +12,6 @@ import PlayerChiclet from './common/PlayerChiclet';
 import * as timeUtils from '../utils/time';
 
 const path = require('path');
-const fs = require('fs');
 
 export default class FileRow extends Component {
   static propTypes = {
@@ -215,12 +214,7 @@ export default class FileRow extends Component {
 
     const file = this.props.file || {};
 
-    const metadata = file.game.getMetadata() || {};
-    const startAt = timeUtils.convertToDateAndTime(metadata.startAt);
-    const getTimeFromFileName = () => timeUtils.filenameToDateAndTime(this.getFileName());
-    const getTimeFromBirthTime = () => timeUtils.convertToDateAndTime(fs.statSync(file.fullPath).birthtime);
-    const startAtDisplay = startAt || getTimeFromFileName() || getTimeFromBirthTime() || 'Unknown';
-
+    const startAtDisplay = timeUtils.fileToDateAndTime(file.game, file.fileName, file.fullPath) || "Unknown";
 
     return <Table.Cell singleLine={true}>{startAtDisplay}</Table.Cell>;
   }
--- a/app/components/FileRow.js
+++ b/app/components/FileRow.js
@@ -12,6 +12,7 @@ import PlayerChiclet from './common/PlayerChiclet';
 import * as timeUtils from '../utils/time';
 
 const path = require('path');
+const fs = require('fs');
 
 export default class FileRow extends Component {
   static propTypes = {
@@ -216,7 +217,7 @@ export default class FileRow extends Component {
 
     const metadata = file.game.getMetadata() || {};
     const startAt = metadata.startAt;
-    const startAtDisplay = timeUtils.convertToDateAndTime(startAt) || 'Unknown';
+    const startAtDisplay = timeUtils.convertToDateAndTime(startAt) || timeUtils.filenameToDateAndTime(this.getFileName()) || timeUtils.convertToDateAndTime(fs.statSync(file.fullPath).birthtime) || 'Unknown';
 
     return <Table.Cell singleLine={true}>{startAtDisplay}</Table.Cell>;
   }
--- a/app/components/FileRow.js
+++ b/app/components/FileRow.js
@@ -216,8 +216,11 @@ export default class FileRow extends Component {
     const file = this.props.file || {};
 
     const metadata = file.game.getMetadata() || {};
-    const startAt = metadata.startAt;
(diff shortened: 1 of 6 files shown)


app/components/FileRow.js
```diff
@@ -216,7 +217,7 @@ export default class FileRow extends Component {
 
     const metadata = file.game.getMetadata() || {};
     const startAt = metadata.startAt;
-    const startAtDisplay = timeUtils.convertToDateAndTime(startAt) || 'Unknown';
+    const startAtDisplay = timeUtils.convertToDateAndTime(startAt) || timeUtils.filenameToDateAndTime(this.getFileName()) || timeUtils.convertToDateAndTime(fs.statSync(file.fullPath).birthtime) || 'Unknown';
```
NikhilNarayana, line 220: Separate each function call into its own variable with a meaningful name

JLaferri: Not sure if that's necessarily a good idea in this case as it will cause them all to actually be evaluated. Probably best to avoid calculating esp the creation time one if it's not needed.

JLaferri: One option to get around this is to make variables like:
`const getFromFileName = () => timeUtils.filenameToDateAndTime(this.getFileName())`

Octopirate1: Unfortunately doesn't work as functions can't be defined within React child functions - thinking of another solution for this though

JLaferri: Honestly I think if you can make a function in timeUtils itself that goes down the list of priority without executing the ones that aren't necessary would be ideal.

JLaferri: In order to get sorting right you may need to calculate the start time at the file level instead so it can be used for sorting properly.

JLaferri:
Perhaps a link will help. My suggestion is to move all of the fallback logic for getting a time to timeUtils function and then after that calling the timeUtils function from here https://github.com/project-slippi/slippi-desktop-app/blob/master/app/actions/fileLoader.js#L232-L237

You can add the time as a new field of the file object and then use that to both display the time in the row as well as sort by.

app/components/FileRow.js
```diff
@@ -214,9 +214,7 @@ export default class FileRow extends Component {
 
     const file = this.props.file || {};
 
-    const metadata = file.game.getMetadata() || {};
-    const startAt = metadata.startAt;
-    const startAtDisplay = timeUtils.convertToDateAndTime(startAt) || 'Unknown';
+    const startAtDisplay = timeUtils.fileToDateAndTime(file.game, file.fileName, file.fullPath) || "Unknown";
```
JLaferri, line 217: Why not get rid of the format arg, always return a moment, and then just use `file.startTime.format('ll · LT')`, making sure that the time is not null, of course.

Octopirate1: I wanted to avoid using moment within places other than timeUtils, especially because app/components/stats/GameProfile.js also uses convertToDateAndTime. If you're ok with using moment in other places then I can change it

app/containers/GameProfilePage.js
```diff
@@ -6,9 +6,11 @@ import { playFile, setStatsGamePage } from "../actions/fileLoader";
 import { dismissError } from "../actions/error";
 
 function mapStateToProps(state) {
+  const index = state.fileLoader.statsGameIndex;
   return {
     store: state.game,
-    statsGameIndex: state.fileLoader.statsGameIndex,
+    statsGameIndex: index,
+    file: state.fileLoader.files[index],
```
JLaferri, line 13: a bit concerned about this. wondering if it would crash if opening a file directly from the file system, for example

Octopirate1: I haven't tested it, but shouldn't that file show up in the ``files`` array?

https://github.com/project-slippi/slippi-desktop-app/blob/b3d0044999f75319c68a12a48ef8ea03a29140ba/app/actions/fileLoader.js#L73-L82

I'm not sure if there's any way to load a file outside of the ``ROOT_FOLDER`` but this should cover the bases if there isn't, no?

JLaferri: There is a way, you can double click an slp file from anywhere. One way to test it is to just `yarn package` your branch and install it. I don't remember how to test it in dev mode though I'm sure I've done it before.

Octopirate1: Just tested it (using files both from root dir and elsewhere). The only thing that happens is that time is set to Unknown again. No crashes, which makes me wonder what's going on. I'll take a deeper look when I figure out how to open console/dev mode when opening a specific file. What I could do is just pass the ``fileLoader`` object and see how to grab the file/attribute that I'm looking for from within the function.

Octopirate1: No idea how to run with a single file in dev mode (tried setting ``DEBUG_PROD`` to true but couldn't open the console, windows cmd piping also doesn't work) or to see the console, which unfortunately means I don't have enough info to fix this bug. However, it isn't crash-inducing.

JLaferri: You might be able to hardcode `argURI` at the following line to point to a replay on your file system temporarily

https://github.com/project-slippi/slippi-desktop-app/blob/master/app/main.dev.js#L374

PR #82, opened 2020-07-16
[PR] [Windows] Elevate permissions before deleting old Dolphin

Kethku:
This PR's goal is to fix the bug where when people install a new version of the desktop app, the old Dolphin cannot be replaced because it was created with elevated permissions itself.

Note: it is ideal if we do not ALSO copy the newer Dolphin with elevated permissions such that we can avoid having to worry about this issue in the future.

This bug would cause the following output when installing 1.5.0:
image

Here are the repro steps:

    Set User Account Control to the default value (one from the top)
    Uninstall the Slippi Desktop App if you have it installed
    Go to the Desktop App AppData folder C:\Users\<user>\AppData\Roaming\Slippi Desktop App and delete the dolphin folder
    Install v1.4.5 of the desktop app: https://github.com/project-slippi/slippi-desktop-app/releases/download/v1.4.5/Slippi-Launcher-Setup-1.4.5.exe
    Install v1.5.0 of the desktop app: https://github.com/project-slippi/slippi-desktop-app/releases/download/v1.5.0/Slippi-Launcher-Setup-1.5.0.exe
    When v1.5.0 opens, you will see the error

Unlike https://github.com/project-slippi/slippi-desktop-app/pull/81, this PR uses https://github.com/jorangreef/sudo-prompt which instead of embedding a binary, creates the script required to elevate permissions on the fly and runs that. This way we don't fall into the trap that electron-sudo did where the binary wouldn't get properly embedded in the webpack bundle.

The way to test is is to run yarn package to build an installer (make sure you have a version of Dolphin in app/dolphin-dev/windows. Then run the same repro steps mentioned above but use your new installer. Kind of obnoxious to test this way but idk another way.


Kethku:
Looks like the formatter was a bit overzealous. Let me know if you would like me to revert the extra fixes it did.

Kethku:
I believe this is working now. I embedded the elevate.exe utility used in the electron-sudo package which can be found here: http://code.kliu.org/misc/elevate/. This gives us the ability to pop a uac dialog without powershell installed and without setting __dirname to true in the webpack node config which as described at the bottom of the readme here: https://github.com/automation-stack/electron-sudo

Kethku:
I ended up swapping to sudo-prompt for the non windows platforms as well because it is more recently updated and maintained. Unfortunately I couldn't get it to work for windows, so its possible we should figure out a method for testing this system on mac and linux.

Kethku:
Lastly its possible the error handling is incorrect in the async wrapper functions. I check if error is null but that may be insufficient for checking if the command actually succeeded. One strategy may be to check if the folder still exists after the elevated delete is finished. That way we don't have to inspect the output of each of the commands, just the result. If folks think this is a worth while change I will add it to the PR.

NikhilNarayana:
```
[2020-07-16 10:24:23.180] [info] 
This version of C:\Program Files\Slippi Launcher\resources\app.asar.unpacked\static\elevate.exe is not compatible with the version of Windows you're running. 
Check your computer's system information and then contact the software publisher.
```
This is what i got when I tried to use it. My UAC was on default.

Kethku:
Thats very weird. Can you give me details about your system? What version of the os are you using?

NikhilNarayana:
Windows 10 Feature Update 2004, @ me on discord if you need extra info.

Kethku:
I don't know of one :( The docs for elevate don't appear to indicate a way to specify the name of the application or icon via command line arguments. I may look into the other utilities highlighted in the elevate.exe url to see what drawbacks or benefits they have and if they support modifying the data in the UAC dialog.

PR #84, opened 2020-07-20
properly store and restore scroll position when viewing recording stats

Kethku:
Fixes https://github.com/project-slippi/slippi-desktop-app/issues/72

Experience changes:
- Scrolling down in the file browser, pressing back, and re-accessing the file browser will no longer scroll a random distance down.
- The position is now properly stored and restored when clicking on a recording's stats and returning to the file loader.

### DIFF
--- a/app/components/FileLoader.js
+++ b/app/components/FileLoader.js
@@ -97,11 +97,6 @@ export default class FileLoader extends Component {
       y: this.refTableScroll.scrollTop,
     });
 
-    // this.props.storeFileLoadState({
-    //   filesToRender: this.state.filesToRender,
-    //   filesOffset: this.state.filesOffset,
-    // });
-
     this.props.dismissError('fileLoader-global');
   }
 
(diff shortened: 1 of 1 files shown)


app/components/FileLoader.js
```diff
@@ -85,11 +97,6 @@ export default class FileLoader extends Component {
       y: this.refTableScroll.scrollTop,
     });
 
-    this.props.storeFileLoadState({
-      filesToRender: this.state.filesToRender,
```
Kethku, line 89: This appears to be left over to when the filesToRender and filesOffset were stored in component state rather than redux. This is no longer needed and causes problems because the values are set to undefined.

NikhilNarayana:
code LGTM, functionality works

I tested with a large set of data and while the load time is longer it gets you to the right spot.

Kethku:
I think this is ready. Is there anything else that needs done here?

PR #85, opened 2020-07-22
Attempt to fix performance issue from relaying

NikhilNarayana:
readPos is useless honestly. We can just send the relay all previous data from the current game when it connects (if it connects during a game) and then send it each piece of new data rather than concating (EXPENSIVE) and splicing (why would you concat and then splice?)

also remove the MIT license because it isn't necessary anymore, the code from this file that needed to be MIT is now LGPL in slippi-js

### DIFF
--- a/app/domain/SlpFileWriter.js
+++ b/app/domain/SlpFileWriter.js
@@ -1,29 +1,3 @@
-/*
-
-MIT License
-
-Copyright (c) 2017 jlaferri
-
-Permission is hereby granted, free of charge, to any person obtaining a copy
-of this software and associated documentation files (the "Software"), to deal
-in the Software without restriction, including without limitation the rights
-to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
-copies of the Software, and to permit persons to whom the Software is
-furnished to do so, subject to the following conditions:
-
-The above copyright notice and this permission notice shall be included in all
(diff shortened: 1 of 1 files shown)


app/domain/SlpFileWriter.js
```diff
@@ -69,6 +69,13 @@ export default class SlpFileWriter extends EventEmitter {
     };
   }
 
+  getFullBuffer() {
+    if (this.currentFile.buffersToConcat.length > 0) 
+      this.currentFile.fullBuffer = Buffer.concat([this.currentFile.fullBuffer, ...this.currentFile.buffersToConcat]);
```
NikhilNarayana, line 74: at this point the question is how fast is this line? I might need to run some tests to get a baseline.

NikhilNarayana: Well i have measurements now and they don't look good for the old method.

https://twitter.com/xMetonym/status/1285995190881148929

NikhilNarayana: Given a Buffer of about 20MB concating with a small 8 byte buffer it takes 8ms, at 40MB it takes 11ms, at 64MB it takes 17ms. This means the old method would continually get worse as you passed the 40ish minute mark. This new method adds a small stutter (a few frames) if the user connects in the middle of the game, but doesn't affect the game otherwise.

PR #88, opened 2020-08-11
Allow showing of games less than 30 seconds

### LINKED ISSUE #78: Add option to change the number of game-seconds that .slp files are filtered by
Right now the browser always filters games shorter than 30sec. This value ought be modifiable. Perhaps something interesting happened during a quick handwarmer.  
I tried implementing it myself, but I have no idea how react works, to be honest.  
https://github.com/kleutzinger/slippi-desktop-app/commit/5867ec5be334823295ee5422dc260a81eeebaf13  
the form to modify the value is mostly nonfuctional/wrong, but other two files are probably close.  

Plj0VB5

vinceau:
Fixes #78 

This PR adds the following:

* a "Click to show all" button which allows the showing of games which have been filtered due to being less than 30 seconds
* the ability to reveal a file in the default file explorer when clicking on the file name

The setting is stored in the local component state which means the option to filter/show is reset back to the default (which is to filter) whenever the component is unmounted.

### Screenshot

screenshot

### DIFF
--- a/app/actions/fileLoader.js
+++ b/app/actions/fileLoader.js
@@ -215,6 +215,7 @@ async function loadFilesInFolder(folderPath) {
       const fullPath = path.join(folderPath, fileName);
       let game = null;
       let hasError = false;
+      let lastFrame = null;
 
       // Pre-load settings here
       try {
@@ -227,7 +228,10 @@ async function loadFilesInFolder(folderPath) {
         }
 
         // Preload metadata
-        game.getMetadata();
+        const metadata = game.getMetadata();
+        if (metadata && metadata.lastFrame !== undefined) {
+          lastFrame = metadata.lastFrame;
+        }
       } catch (err) {
         console.log(`Failed to parse file: ${fullPath}`);
         console.log(err);
@@ -242,6 +246,7 @@ async function loadFilesInFolder(folderPath) {
         startTime: startTime,
         game: game,
         hasError: hasError,
+        lastFrame: lastFrame,
       };
     })
   ));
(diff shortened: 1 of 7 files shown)


JLaferri:
Trying to think of text that seems more actionable. Like "Click to show all"?

vinceau:
> 
> 
> Trying to think of text that seems more actionable. Like "Click to show all"?

How's "Show all replays"? Isn't "Show" already an action?

NikhilNarayana:
I would like to see it as clear button, I'm worried users won't realize that text is interactable

vinceau:
> 
> 
> I would like to see it as an obvious button, I'm worried users won't realize that text is interactable

I intentionally didn't want to spend too much time on the styling since I think you guys would have a better idea of what you want the button to look like and what kinda aesthetics you're after.

JLaferri:
Yeah I think what Nikki mentioned is why I wanted to add the word "Click". But it's a pretty small thing, I'm content enough pushing it as is, UX-wise. I trust you enough to not have to go through the code too deeply lol

vinceau:
Changed it to "Click to show all" but I haven't changed the styles. Hopefully it's still obvious enough to users.

JLaferri:
What does it look like after you click the button in the case where you:
1. Didn't have corrupt replays
2. Did have corrupt replays

vinceau:
> What does it look like after you click the button in the case where you:
> 
>     1. Didn't have corrupt replays
> 
>     2. Did have corrupt replays

In both cases, the entire banner is hidden after the user clicks "Click to show all". In case number 2 we probably want to keep the banner to indicate the number of corrupted replays but the text copy would need to change.

vinceau:
Updated the banner to show this in the case that the button has been clicked but there are still corrupted files which were filtered.

screenshot

NikhilNarayana:
LGTM

PR #91, opened 2020-08-30
Add Delete Button to selection menu

b3nd3r-ssbm:
deleteButton
There is now a "delete" button when you select files.

### DIFF
--- a/app/components/FileLoader.js
+++ b/app/components/FileLoader.js
@@ -34,7 +34,8 @@ export default class FileLoader extends Component {
     storeScrollPosition: PropTypes.func.isRequired,
     storeFileLoadState: PropTypes.func.isRequired,
     setStatsGamePage: PropTypes.func.isRequired,
-
+    deleteSelections:  PropTypes.func.isRequired,
+    
     // error actions
     dismissError: PropTypes.func.isRequired,
 
@@ -413,6 +414,14 @@ export default class FileLoader extends Component {
       </Table>
     );
   }
+  
+  deleteSelections = () => {
+    this.props.deleteSelections(this.state.selections);
+    this.setState({
+      selections: [],
+    });
+    this.renderMain();
+  }
 
   renderQueueButtons() {
     if (this.state.selections.length === 0) {
@@ -427,6 +436,10 @@ export default class FileLoader extends Component {
         <Button onClick={this.queueClear}>
           <Icon name="dont" />
           Clear
+        </Button>
+        <Button onClick={this.deleteSelections}>
+          <Icon name="trash alternate outline" />
+            Delete
         </Button>
       </div>
     );
--- a/app/actions/fileLoader.js
+++ b/app/actions/fileLoader.js
@@ -3,6 +3,7 @@ import fs from 'fs';
 import path from 'path';
 import SlippiGame from '@slippi/slippi-js';
 import * as timeUtils from '../utils/time';
+import { shell } from 'electron';
 
 import { displayError } from './error';
 import { gameProfileLoad } from './game';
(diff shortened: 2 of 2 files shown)


app/components/FileRow.js
```diff
@@ -258,13 +255,16 @@ export default class FileRow extends Component {
...
   render() {
-    return (
-      <Table.Row>
-        {this.generateSelectCell()}
-        {this.generateDetailsCell()}
-        {this.generateStartTimeCell()}
-        {this.generateOptionsCell()}
-      </Table.Row>
-    );
+    if(fs.existsSync(this.props.file.fullPath)){
```
vinceau, line 258: Just my two cents:
* I think it's better to actually remove the file from the entire file list rather than check for file existence before rendering. The rendering should expect the file to exist and just render the component rather than worry about it not existing.
* It's probably a good idea to use [`shell.moveItemToTrash`](https://www.electronjs.org/docs/api/shell#shellmoveitemtotrashfullpath-deleteonfail) instead of `fs.unlink` since that way the files can be recovered from the Recycle bin if they so wish.

b3nd3r-ssbm: Ok, I'll try that out

b3nd3r-ssbm: I did it

app/components/FileLoader.js
```diff
@@ -20,6 +20,8 @@ import FolderBrowser from './common/FolderBrowser';
 import PageWrapper from './PageWrapper';
 import Scroller from './common/Scroller';
 
+const shell = require('electron').shell;
```
vinceau, line 23:
```suggestion
import { shell } from 'electron';
```

It's probably best to avoid mixing ES6 and require syntax.

app/components/FileLoader.js
```diff
@@ -413,13 +415,28 @@ export default class FileLoader extends Component {
...
+  
+  deleteSelections = () => {
+    const store=this.props.store || {};
+	let filesToRender=_.get(store, ['fileLoadState', 'filesToRender']);
+	let i=0;
+    for(i=0;i<this.state.selections.length;i++){
+      shell.moveItemToTrash(this.state.selections[i].fullPath);
+	  filesToRender.splice(filesToRender.indexOf(this.state.selections[i]),1);
+	  _.set(store, ['fileLoadState', 'filesToRender'],filesToRender);
+    }
```
vinceau, line 427: I think it would improve readability if you used either the `for of` syntax or the `.forEach()` list operator for iterating through files. Also using `_.set()` is not the correct way to dispatch changes to the Redux store since I believe `_.set()` just modifies the object in place. Additionally, it would probably be more performant if you updated the Redux store once, after all the files have been deleted, rather than updating the store for each deleted file.

vinceau (changes requested):

app/components/FileLoader.js
b3nd3r-ssbm: fair point

JLaferri:
Text should just say "delete"

b3nd3r-ssbm:
> 
> 
> Text should just say "Delete"

It now does, I'll update the picture when I finish up with Vince's suggested changes :)

app/components/FileLoader.js
```diff
@@ -413,6 +415,20 @@ export default class FileLoader extends Component {
       </Table>
     );
   }
+  
+  deleteSelections = () => {
```
JLaferri, line 419: Actions like this should be done in redux. With the current logic, if anything else ever updates the redux stores, I'm pretty sure all those deleted files will come back. Currently I'm not too sure if there's an action that will actually do that but one could be added at any time.

Look at how redux actions are fired in other components and put this handler as an action.

b3nd3r-ssbm: OK, I'll learn redux lol

vinceau (approved):

PR #94, opened 2020-10-22
Fix unfiltering

NikhilNarayana:
#88 had an issue where game profiles would fail to load the right game if there were hidden files because it was referencing the list of all files. This PR fixes that by saving the unfiltered and filtered list in the store and selecting the right list to check.


app/actions/fileLoader.js
```diff
@@ -194,22 +205,31 @@ export function setStatsGamePage(index) {
...
-          fileLoadState: tempStore,
-        },
-      });
-    };
+  return (dispatch, getState) => {
+    const tempStore = getState().fileLoader.fileLoadState;
+    const filesToRender = tempStore.filesToRender;
+    _.each(selections, (selection, i) => {
+      shell.moveItemToTrash(selection.fullPath);
+      filesToRender.splice(i, 1);
```
vinceau, line 213: A few comments regarding this, we should probably almost always avoid using `splice()` since this modifies the array in place. Rather, we should probably either use `slice()` or even better, instead of constantly modifying the `filesToRender` file list with each iteration, we could probably just create a new list once, after moving all the items to trash.

Perhaps something like:
```
// do the iteration at the beginning
// then after that update the files list
filesToRender = tempStore.filesToRender.filter(() => /* something to filter out the selection */);
```

app/components/FileLoader.js
```diff
@@ -22,7 +22,6 @@ import Scroller from './common/Scroller';
 
 const GAME_BATCH_SIZE = 50;
 const MIN_GAME_LENGTH_SECONDS = 30;
```
vinceau, line 24: What do you think about having this value returned from the `fileLoader`? Otherwise there will be two sources of truth if we ever change that min game length value.

vinceau (approved):
LGTM

PR #97, opened 2020-11-18
Add Match Filtering to Replay Browser

DanielKilgallon:
Hey!

I've been working on adding four types of filtering to the main replay view, Character, Player, Stage, and Filename. Take a look at the code, I've tried to optimize it to reduce the slowdown when filtering a large number of matches

Here's a demo of it in action:
Slippi-Desktop-App-2020-11-18-16


seanlaguna:
I tested this and found one bug. When I click on the info icon for a filtered match it doesn't take me to the correct info page. Clicking the "play" button plays the correct replay, though.

aadunn:
Happy to see someone hop onto this. I've thought about it a few times when going back through my replays!

NikhilNarayana:
only accepting PRs for the rewrite at this point

PR #105, opened 2020-11-28
Make Conversion Timestamps Play Replay from that conversion

b3nd3r-ssbm:
ContentIndolentBudgie-size_restricted
sorry for the random black frames, OBS was acting strangely whenever dolphin opened during recording...

### DIFF
--- a/app/actions/fileLoader.js
+++ b/app/actions/fileLoader.js
@@ -1,7 +1,7 @@
 import _ from 'lodash';
 import fs from 'fs';
 import path from 'path';
-import SlippiGame from '@slippi/slippi-js';
+import { SlippiGame, Frames } from '@slippi/slippi-js';
 import { shell } from 'electron';
 import * as timeUtils from '../utils/time';
 
@@ -145,7 +145,7 @@ export function storeFileLoadState(fileLoadState) {
   };
 }
 
-export function playFile(file) {
+export function playFile(file, startFrame=Frames.FIRST) {
   return async (dispatch, getState) => {
     const filePath = file.fullPath;
     if (!filePath) {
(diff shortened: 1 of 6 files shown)


app/actions/fileLoader.js
```diff
@@ -165,6 +165,25 @@ export function playFile(file) {
   };
 }
 
+export function playFileFromFrame(filePath, startFrame) {
+  return async (dispatch, getState) => {
+    if (!filePath) {
+      // TODO: Maybe show error message
+      return;
+    }
```
vinceau, line 173: Probably don't need this.

app/domain/DolphinManager.js
```diff
@@ -174,6 +174,21 @@ export default class DolphinManager extends EventEmitter {
...
+      replay: filePath,
+      startFrame: startingFrame,
+      isRealTimeMode: this.settings.isRealTimeMode || false,
+      commandId: uniqueId, // Indicates to Dolphin to play new replay
+    });
+
+    await this.writeCommFile(jsonString);
+    await this.runDolphin(true);
+  }
+
```
vinceau, line 191: Can you just have the `startFrame` be an optional parameter in the `playFile()` method? I don't think it makes sense to essentially write the same function twice.

vinceau (changes requested):

app/domain/DolphinManager.js
b3nd3r-ssbm: great idea, I should do that

app/actions/fileLoader.js
```diff
@@ -165,6 +165,25 @@ export function playFile(file) {
   };
 }
 
+export function playFileFromFrame(filePath, startFrame) {
```
vinceau, line 168: This entire function shouldn't be needed anymore. You should update the `playFile()` action with the new interface and remove pretty much all references to `playFileFromFrame()`.

vinceau (changes requested):

app/actions/fileLoader.js
b3nd3r-ssbm: but `playFile()` requires a file, not a string to a file path. How do you suggest I go about either getting a `file` object from the file path or making the function support file paths?

NikhilNarayana (approved):

PR #109, opened 2020-12-02
Global Game Stats

alvarosevilla95:
This PR changes how game loading works. Instead of only parsing the base .slp file on FileLoader and deferring the full stats parsing to the game stats page, it creates a stats subdirectory under the replay dir where it dumps the metadata,settings,stats of every file in json format. This causes a longer initial replay list load (around 10minutes on my laptop for 2.5k games), but allows for fast loading in subsequent runs.

Having the global stats data available, we can enrich the main replay list, for example adding a final stock count image for each player. I've also added a new Route /player, which can be accessed by clicking on a players tag on the replay list, and displays a variety of global stats for that player.

Demo: https://drive.google.com/file/d/1wyAf53bx8sBGntDSirVEc_0PUF-wQ1mN/view?usp=sharing


alvarosevilla95:
I'm not very familiar with es-lint. I'll push a cleanup for the errors. I'll clean up and rebase removing the readme stuff later

PR #111, opened 2020-12-08
Bundle Netplay Dolphin

alvarosevilla95:
Adds a new button to launch netplay dolphin running melee.
To achieve this a netplay dolphin is bundled with the config logic
replicated from the playback dolphin


NikhilNarayana:
not needed with the rewrite

PR #113, opened 2020-12-20
Only validate when necessary

NikhilNarayana:
If the file modification time hasn't changed then we can naively skip validating again.

### DIFF
--- a/src/renderer/store/settings/index.ts
+++ b/src/renderer/store/settings/index.ts
@@ -73,7 +73,7 @@ export const useSettings = create<StoreState & StoreReducers>((set, get) => ({
       })
     );
   },
-  verifyIsoPath: async (isoPath, shouldSetPath) => {
+  verifyIsoPath: async (isoPath) => {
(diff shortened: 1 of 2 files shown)


src/renderer/store/settings/index.ts
```diff
@@ -55,28 +59,43 @@ export const useSettings = create<StoreState & StoreReducers>((set, get) => ({
...
     // Indicate that we're loading
     set({
       verifyingIso: true,
       validIsoPath: null,
     });
 
     try {
-      const res = await verifyISO(isoPath);
-      set({ validIsoPath: res.valid });
+      const storedModTime = get().settings.isoModTime;
```
vinceau, line 84: If you're calling `get()` a lot, something that might be easier and potentially more readable is using object destructuring syntax. e.g

```typescript
const { settings, setIsoPath, setIsoModTime } = get();
```

vinceau (approved):

PR #115, opened 2021-01-01
Stats page!

Randypaquette13:
Added components previously found at /app/components/stats

Currently the Launch replay button doesn't work although that may be outside the scope of this task let me know if I should look into getting that functionality to work.


src/renderer/containers/ReplayFileStats/EdgeGuardTable.tsx
```diff
@@ -0,0 +1,101 @@
+import { ComboType, StatsType } from "@slippi/slippi-js/dist/stats/common";
```
vinceau, line 1: I think it's fine to remove this file from source control since it's not used.

src/renderer/containers/ReplayFileStats/OverallTable.tsx
```diff
@@ -0,0 +1,403 @@
...
+
+const columnCount = 5; // Unfortunately there is no way to specify a col span of "all" max cols there will be is 5
+
+export interface OverallTableProps {
+  file: FileResult;
+  stats: StatsType | null;
+}
+
+export const OverallTable: React.FC<OverallTableProps> = ({ file, stats }) => {
+  if (!stats) return <div>An Error Occurred!</div>;
```
vinceau, line 17: `stats` should not be nullable. If `stats` is null, the parent should render an error message and the child components should never have to handle the null case.

src/renderer/containers/ReplayFileStats/index.tsx
```diff
@@ -11,20 +13,15 @@ export interface ReplayFileStatsProps {
...
+export const ReplayFileStats: React.FC<ReplayFileStatsProps> = (props) => {
+  // const { file, index, total, onNext, onPrev, onClose } = props;
+
   const loading = useReplays((store) => store.selectedFile.loading);
   const error = useReplays((store) => store.selectedFile.error);
   const gameStats = useReplays((store) => store.selectedFile.gameStats);
   return (
     <div>
-      <div>{file.name}</div>
+      {/* <div>{file.name}</div>
```
vinceau, line 24: The `onClose` event needs to be handled rather than just commented out. Since we no longer have a "back" button, we need a way to go back from the stats page to the replay file list.

src/renderer/containers/ReplayFileStats/PunishTable.tsx
```diff
@@ -0,0 +1,292 @@
...
+      return i + "st";
+    }
+    if (j == 2 && k != 12) {
+      return i + "nd";
+    }
+    if (j == 3 && k != 13) {
+      return i + "rd";
+    }
+    return i + "th";
+  };
```
vinceau, line 37: This should probably be an exported function in like `lib` somewhere rather than inside the component.

src/renderer/containers/ReplayFileStats/PunishTable.tsx
```diff
@@ -0,0 +1,292 @@
...
+  stats: StatsType | null;
+  playerIndex: number;
+}
+
+export const PunishTable: React.FC<PunishTableProps> = ({
+  file,
+  stats,
+  playerIndex,
+}) => {
+  if (!stats) return <div>An Error Occurred!</div>;
```
vinceau, line 23: Again, as mentioned before, child components shouldn't have to handle the case where `stats` is null. It should be handled by the parent.

vinceau (changes requested):
For starters, amazing work! Overall very happy with the quality of the code, just have a few places that need fixing up as well as a DOM nesting issue as seen in the screenshot below.

dom nesting

src/renderer/containers/ReplayFileStats/GameProfile.tsx
```diff
@@ -0,0 +1,276 @@
...
+  //         <div style={{ display: "inline-block", verticalAlign: "top" }}>
+  //           <EdgeGuardTable file={file} stats={stats} playerIndex={0} />
+  //         </div>
+  //         <div style={{ display: "inline-block", verticalAlign: "top" }}>
+  //           <EdgeGuardTable file={file} stats={stats} playerIndex={1} />
+  //         </div>
+  //       </div>
+  //     </div>
+  //   );
+  // };
```
vinceau, line 245: This should be removed too.

vinceau (approved):

Issue #112, opened 2020-12-09
Rewrite it entirely


NikhilNarayana:
The desktop app is long overdue for a rewrite. This issue is just so we can keep track of the overall work to be done. Work is being done in https://github.com/project-slippi/slippi-desktop-app/tree/refactor/rewrite and `rewrite/*` branches. Always make a branch off of `refactor/rewrite` with the name `rewrite/<feature name>`.

The rewrite is now open to outside contributions. However I suggest you discuss code design with me and Vince before implementing something so you don't waste time on something we don't like.

### Deferred for now
* Gecko code disabling (Nikki)
* Replay caching (#126)
* Global Stats (#125)
* Replay grouping
  * [Like this](https://user-images.githubusercontent.com/6331403/110841444-ba591580-825a-11eb-8a0a-7e1c6cb1df23.png)
* Friends list
  * Should be able to add friends manually or from the replay browser.
  * Should let you invite friends to play some games (no chat though).
  * Will likely require Fizzi supervision.
* Visualizer
* Show patch notes after an update (#132)
* Multiple top level folders (#131)
* Install GCC Adapter drivers on install of desktop app or with the click of a button
  * I don't want this merged in until after the first release of the new desktop app. The issue is that i have no clue if antivirus software programs will complain about what it is doing.
  * To do this all you need to do is run the simple installer found here: https://github.com/project-slippi/libwdi/releases/tag/slippi-v1

### Done
* Logging in with slippi.gg credentials
* ~~Big~~ Medium play button
* Downloading the latest netplay dolphin
* Launching netplay Dolphin
* Replay browser (#117)
* Game Profile (#115)
* Download latest playback dolphin (#120)
* Auto update dolphin by comparing version with release (#120)
* DolphinManager (#121)
  * manages all instances of dolphin (netplay, playback, spectate, config) and launches them from the main thread
* Download dolphin on main thread (#124)
* News feed (#129)
* Dolphin Settings page (#130)
* *Console/Broadcasting (rewrite/broadcast-spectator branch)* (#148)
  * basic notes: https://imgur.com/a/QVGQpEq
* Prompt for local play when not logged in
* Finish updating styles to match designs
* New icon designs
* Prompt and quick start page for when user hasn't created online connect code
* Console mirroring and autoswitcher (rewrite/mirroring)
* merge master changes like v6 of slippi-js, new stats etc.
* Test playback + netplay dolphin installing and auto-updating on all platforms
* QA testing on mac and linux
[...]

PR #118, opened 2021-01-23
POC: Slippi visualiser on stats

Fcornaire:
# This is a Proof of concept and not merge ready !

First time contribution here, i added the visualiser used in the preview on slippi.gg/tournament in the stats page.
visualiser

It only appear with large combo/conversion (Red heart) and play only on focus.
This PR is not merge ready. I want feedback and discussion about my implementation since there is still a lot other thing to fix mostly in the [react wrapper](https://github.com/Fcornaire/react-slippi-visualiser) that i made.

Basically, i made a wrapper that take a replay (parser) and send it to the slippi-visualiser original lib throught react.
The desktop app only need a tweak about Jquery and to load all the animations in the html.
(i also had to update react since i use hook in the wrapper 😅)

What's do you think about this ?

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -19,7 +19,11 @@
     "generator-star-spacing": "off",
     "import/no-unresolved": "error",
     "import/no-extraneous-dependencies": "off",
-    "indent": ["error", 2, { "SwitchCase": 0 }],
+    "indent": [
+      "error",
+      2,
+      { "SwitchCase": 0, "ignoredNodes": ["TemplateLiteral"] }
+    ],
     "jsx-a11y/label-has-associated-control": "off",
     "jsx-a11y/label-has-for": "off",
     "jsx-a11y/click-events-have-key-events": "off",
--- a/app/app.html
+++ b/app/app.html
@@ -1,7 +1,7 @@
 <!DOCTYPE html>
 <html>
   <head>
-    <meta charset="utf-8">
+    <meta charset="utf-8" />
     <title>Slippi Desktop App</title>
     <script>
       (function() {
@@ -12,7 +12,13 @@
           // HACK: Writing the script path should be done with webpack
           document.getElementsByTagName('head')[0].appendChild(link);
         }
-      }());
+      })();
+    </script>
+    <script>
+      require('react-slippi-visualiser/dist/animations');
+    </script>
+    <script>
+      window.$ = window.jQuery = require('jquery');
     </script>
   </head>
   <body>
@@ -30,7 +36,7 @@
         // Dynamically insert the bundled app script in the renderer process
         const port = process.env.PORT || 1212;
         scripts.push(
-          (process.env.HOT)
+          process.env.HOT
             ? 'http://localhost:' + port + '/dist/renderer.dev.js'
             : './dist/renderer.prod.js'
         );
--- a/app/components/stats/PunishesTable.js
+++ b/app/components/stats/PunishesTable.js
@@ -4,6 +4,7 @@ import React, { Component } from 'react';
 import PropTypes from 'prop-types';
 import { Table, Image, Icon } from 'semantic-ui-react';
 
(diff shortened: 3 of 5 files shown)


JLaferri:
I think it's a cool idea but right now it's way too big.

If anything maybe it should be behind a click and open in a dialog or something. But also we're currently working on a refactor of the desktop app so im not sure whether something like this would get merged.

Maybe @vinceau has thoughts.

Fcornaire:
oh yeah ofc, not a problem if it's not implemented right now or never, was just testing to see how it will turn out.
I also heard about the rewrite but i thought it's being made internaly ?

like a said it's just a POC, i thought about the dialog but my problem with it was like you have and extra step to show something that should be quick to see, since the visualiser is not a replacement of the replay itself.
i think it's cool seeing quick highlights but obviously it's not final or anything, so 😅

vinceau:
@Fcornaire This is a very cool concept! Nice work! There's definitely benefit to having the game show up in the desktop app and without having to load it through Dolphin. I agree with @JLaferri that it probably should be hidden behind a click in a dialog, more like how slippi.gg shows the game previews. We should also clearly distinguish between playing it back in browser verus playing it back in Dolphin (added in #105).

Some comments regarding the code and implementation:

1. **The Rewrite**. For starters, I don't think we should be accepting anymore PRs into the "legacy" desktop app. As per #112, we'll be rewriting the desktop app in Typescript so new features should have PRs pointed to the [`refactor/rewrite`](https://github.com/project-slippi/slippi-desktop-app/tree/refactor/rewrite). Ideally imported libraries are also written in Typescript or provide `.d.ts` typings out of the box.
2. **Global imports**. I'm not a big fan of having to globally import jQuery and all the animations required for getting the game preview to work. Since you've added them to the root `app.html` they will always be loaded regardless of page, and regardless if the game is shown. Ideally we should not have to do any global imports (including global CSS) to use the `react-slippi-visualiser` component.
3. **Depending on jQuery is not great...**. React is already a fairly big library. jQuery is an even bigger, chonkier boi. So use of jQuery should be avoided. Additionally, since jQuery modifies the DOM outside of React's shadow-DOM we lose out on the benefits of using React in the first place. Ideally we should not have to depend on jQuery at all and instead rely on passing React refs for DOM manipulation.
4. **Melee light/visualizer is still largely incomplete**. Out of the 26 character cast, only 5 characters are supported (<20%). By including all the animations (~30MB) in a [compiled and minified format](https://github.com/Fcornaire/react-slippi-visualiser/blob/master/public/animations.js) it makes expanding the number of supported characters very difficult. Ideally the animations are included in the React library itself and can be individually lazy-loaded when needed.

I've messed around a bit with [react-slp-viewer](https://github.com/vinceau/react-slp-viewer) to try and fix some of these issues. It supports Typescript, does not depend on global imports, and lazy-loads the animations on demand. Let me know what you think about these comments and ideas, and if you would like to collaborate on this together.

Fcornaire:
@vinceau & @JLaferri  Thanks for the feedback. 
You mostly answered the thing i thought needed some rework:
 1. I'm gonna wait for now since this branch [refactor/rewrite](https://github.com/project-slippi/slippi-desktop-app/tree/refactor/rewrite) probably gonna change a lot, so i will likely wait for the full re write since this feature is not mandatory (i could also help on the refacto but i feel like my help is not really needed 😅 )
 2. Yup , didn't  liked to do that but it's was mandatory for the visualiser to work. Again it's just a POC using what is already there to see how thing will work
 3. the ideal will be no JQuery, the original lib need rework
 4. I actually searched the source but only found the compiled version 😥. Was not a fan of loading all that in the html  but it was the only solution i found for now at least. I think ideally the original lib should manage that
 
I'm gonna look into your implementation  [react-slp-viewer](https://github.com/vinceau/react-slp-viewer) and try to go from there.

I'm also closing this PR since it's useless now

PR #120, opened 2021-01-30
Auto download latest Playback Dolphin and auto update both dolphins

NikhilNarayana:
~~No version checking logic yet (playback doesn't support it yet), should be added before release since 2.3.0 should be the norm and that will require a playback update~~


NikhilNarayana:
added version checking cause it made sense to just do now, however this change kinda forces us to use node 12+ cause I needed spawnSync. Nothing seems to break, so it should be ok

src/renderer/lib/directories.ts
```diff
@@ -15,12 +15,15 @@ export function getDefaultRootSlpPath(): string {
   return path.join(root, "Slippi");
 }
 
-export async function findDolphinExecutable(): Promise<string> {
+export async function findDolphinExecutable(
+  type: DolphinType
+): Promise<string> {
   // Make sure the directory actually exists
-  await fs.ensureDir(NETPLAY_PATH);
+  const dolphin_path = path.join(remote.app.getPath("userData"), type);
```
vinceau, line 22: We should probably should stick with camelCase for variable names.

```suggestion
  const dolphinPath = path.join(remote.app.getPath("userData"), type);
```

src/renderer/lib/downloadDolphin.ts
```diff
@@ -1,44 +1,84 @@
...
 ): Promise<void> {
   try {
-    await findDolphinExecutable();
-    log("Found existing Dolphin executable.");
+    await findDolphinExecutable(type);
+    log(`Found existing ${type} Dolphin executable.`);
+    log(`Checking if we need to update`);
+    const data = await getLatestReleaseData(type);
+    const latest_version = data.tag_name?.substring(1);
+    const is_outdated = await compareDolphinVersion(type, latest_version);
```
vinceau, line 22: You don't need to manually split out the `v` from the semver string. When using the `semver` package, it's smart enough to compare strings with or without the `v` prefix and the `compareDolphinVersion` calls `lt` from the `semver` package.

Also something something camelCase.

```suggestion
    const latestVersion = data.tag_name?.substring(1);
    const isOutdated = await compareDolphinVersion(type, latest_version);
```

src/renderer/lib/downloadDolphin.ts
```diff
@@ -79,27 +120,38 @@ async function downloadLatestNetplay(
...
       const zip = new AdmZip(assetPath);
-      log(`Extracting to: ${extractToLocation}, and renaming to netplay`);
-      zip.extractAllTo(extractToLocation, true);
-      const oldPath = path.join(extractToLocation, "FM-Slippi");
-      const newPath = NETPLAY_PATH;
-      if (await fs.pathExists(newPath)) {
-        log(`${newPath} already exists. Deleting...`);
-        await fs.remove(newPath);
+      const already_installed = await fs.pathExists(dolphin_path);
+      if (already_installed) {
```
vinceau, line 139: 🐫

```suggestion
      const alreadyInstalled = await fs.pathExists(dolphin_path);
      if (alreadyInstalled) {
```

vinceau (changes requested):
Only some minor changes.

src/renderer/lib/downloadDolphin.ts
NikhilNarayana: woah, that's nice

vinceau (approved):
LGTM!

vinceau (approved):

vinceau (approved):

src/renderer/lib/downloadDolphin.ts
```diff
@@ -126,70 +126,43 @@ async function installDolphin(
...
-        await fs.move(
-          dolphinPath,
-          path.join(remote.app.getPath("userData"), type + "_old")
-        );
-      }
+      const backupLocation = path.join(
+        remote.app.getPath("userData"),
+        type + "_old"
+      );
+
```
vinceau, line 139: Probably don't need to declare this twice.

vinceau (changes requested):

vinceau (approved):

PR #121, opened 2021-01-31
Add DolphinManager and move functionality to main thread

NikhilNarayana:
This adds a DolphinManager class that mimics stable except it is a singleton that manages any number of dolphin instances at once. At the moment it only supports playback and netplay, config can be added once we add the necessary UI. All of this is also done on the main thread to reduce UI issues in the future when we add more complex functionality like relay.


src/renderer/lib/startDolphin.ts
```diff
@@ -0,0 +1,72 @@
+import { DolphinType } from "./directories";
+import { remote } from "electron";
+import { randomBytes } from "crypto";
+import { assertDolphinInstallation, openDolphin } from "./downloadDolphin";
+import { assertPlayKey } from "./playkey";
+import * as fs from "fs-extra";
+
+import path from "path";
+
+export interface ReplayComm {
```
vinceau, line 10: Can we rename this to not be short-hand? Something like `ReplayCommunication` or something? I feel like exported interfaces and function names shouldn't be shortened.

src/renderer/lib/startDolphin.ts
```diff
@@ -0,0 +1,73 @@
...
+
+  log("Starting replay...");
+  const dolphin = await openDolphin(DolphinType.PLAYBACK, ["-i", replayCommPath, "-b", "-e", meleeFile]);
+  dolphin.on("close", () => {
+    log(`deleting ${replayCommPath}`);
+    fs.unlink(replayCommPath);
+  });
+}
+
+async function genCommFilePath(): Promise<string> {
```
vinceau, line 66: Is `generateCommunicationFilePath` too long? 😅

vinceau (changes requested):

src/renderer/lib/startDolphin.ts
NikhilNarayana: i just copied it from stable. We can rename if we want

src/renderer/lib/startDolphin.ts
NikhilNarayana: gotcha

vinceau (approved):
LGTM! Merge whenever you're happy with it.

src/main/dolphinManager.ts
```diff
@@ -0,0 +1,217 @@
+import { ChildProcessWithoutNullStreams, spawn } from "child_process";
+import { randomBytes } from "crypto";
+import { app } from "electron";
+import electronSettings from "electron-settings";
+import { EventEmitter } from "events";
+import * as fs from "fs-extra";
+import { find, remove } from "lodash";
+import path from "path";
+
+import { DolphinType } from "../renderer/lib/directories";
```
vinceau, line 10: It's important that we don't mix up things that are imported in `main` and things that are imported in `renderer`. If it's something that's used by both, we should move it to `common`.

src/renderer/lib/startDolphin.ts
```diff
@@ -0,0 +1,73 @@
+import { randomBytes } from "crypto";
```
vinceau, line 1: I don't believe we're using this file anymore so it should be removed.

vinceau (changes requested):

src/main/dolphinManager.ts
```diff
@@ -0,0 +1,216 @@
...
+export interface ReplayQueueItem {
+  path: string;
+  startFrame?: number;
+  endFrame?: number;
+  gameStartAt?: string;
+  gameStation?: string;
+}
+
+interface DolphinInstance {
+  dolphinUseType: "playback" | "spectate" | "config" | "netplay";
```
vinceau, line 38: Maybe have this as an exported type in `common`?
e.g.
```
export type DolphinType = "playback" | "spectate" | "config" | "netplay";
```
so then you can use it elsewhere to like in line 53.

Or alternatively, even better is probably having this as an enum, like:
```
export enum DolphinType {
  PLAYBACK = "PLAYBACK"
  SPECTATE = "SPECTATE",
  CONFIG = "CONFIG",
  NETPLAY = "NETPLAY",
}
```

src/main/dolphinManager.ts
```diff
@@ -0,0 +1,216 @@
...
+  const commFileName = `slippi-${dolphinUseType}-${uniqueId}.txt`;
+  const commFileFullPath = path.join(tmpDir, commFileName);
+
+  return commFileFullPath;
+}
+
+// DolphinManager should be in control of all dolphin instances that get opened for actual use.
+// This includes playing netplay, viewing replays, watching broadcasts (spectating), and configuring Dolphin.
+// The openDolphin function is exported so other parts of the app can do some checks.
+export class DolphinManager extends EventEmitter {
```
vinceau, line 66: Is there a reason why we extend `EventEmitter` but we don't actually emit any events with `this.emit()`?

src/main/dolphinManager.ts
```diff
@@ -0,0 +1,216 @@
...
+        dolphin.on("close", () => {
+          this.dolphinInstances[configureType] = null;
+        });
+
+        break;
+      }
+    }
+  }
+
+  private async startDolphin(
```
vinceau, line 200: I think this could not be attached to the class since it's not public and not instance-specific. e.g. just having in the file somewhere
```
function async startDolphin(...) {
....
}
```

vinceau (changes requested):

vinceau (approved):
LGTM!

PR #123, opened 2021-02-10
Change broadcasting start time to a locale time string

qtran09:
When broadcasting, the start time is converted into a string using JSON.stringify, which turns the formatting into an ISO Date-Time format. Using a locale time string is more readable in this case.

### When using JSON.stringify(date): 
When using JSON.stringify

### When using date.toLocaleString(): 
When using date.toLocaleString()


NikhilNarayana:
Not accepting further contributions to the legacy desktop app.

PR #124, opened 2021-02-12
Move downloadDolphin to main thread

NikhilNarayana:
We should figure out a better way to handle knowing that the logs are done coming in cause this doesn't scale if we need to do more on init.


src/main/downloadDolphin.ts
```diff
@@ -1,13 +1,21 @@
...
 import { fileExists } from "common/utils";
-import { remote } from "electron";
+import { fetch } from "cross-fetch";
+import { app, BrowserWindow } from "electron";
+import { download } from "electron-dl";
 import * as fs from "fs-extra";
 import path from "path";
 import { lt } from "semver";
 
+export function sendToRenderer(message: string, channel = "downloadDolphinLog"): void {
```
vinceau, line 12: We probably shouldn't export this, at least not here.

src/main/downloadDolphin.ts
```diff
@@ -115,8 +132,8 @@ async function installDolphin(
   assetPath: string,
   log: (message: string) => void = console.log,
 ) {
-  const dolphinPath = path.join(remote.app.getPath("userData"), type);
-  const backupLocation = path.join(remote.app.getPath("userData"), type + "_old");
+  const dolphinPath = path.join(app.getPath("userData"), type);
+  const backupLocation = path.join(app.getPath("userData"), type + "_old");
```
vinceau, line 136: Why not something like this?
```suggestion
  const dolphinPath = path.join(app.getPath("userData"), type);
  const backupLocation = dolphinPath + "_old";
```

src/renderer/store/app/index.ts
```diff
@@ -70,7 +68,6 @@ export const useApp = create<StoreState & StoreReducers>((set, get) => ({
     }
 
     await Promise.all(promises);
-    set({ initialized: true });
```
vinceau, line 73: This seems like a problem if we only complete initialization on `downloadDolphinFinished`. What happens if you disconnect from the internet and start the app? The download would never complete and initialized would never be set.

vinceau (changes requested):

src/renderer/store/app/index.ts
vinceau: I think one solution is to have the `downloadDolphinFinished` receive an optional error object. Then send that error object along with the `downloadDolphinFinished` event if an error occurs.

vinceau (approved):

PR #126, opened 2021-02-25
Persist replay stats data to sqlite for faster access

alvarosevilla95:
This PR implements a simple sqlite database to cache replay data for faster access. 
The replay loading process occurs as follows:
* Existing file names are retrieved from the db
* Folder is scanned for slp files
* All files in the db and not in the folder are deleted from the db
* Files in the folder and not in the db are parsed in worker threads and stored in the db.
* Replay data is loaded from the database

The `loadFolder` endpoint wraps this entire process and can be provided a callback to get load progress. This callback is used to update the loading screen.

At the moment this process is async but files are returned in bulk at the end, but youcould easily change the callback logic to instead update files continuosly. However that involves some ux issues like handling the user clicking away and with the multithreaded loading I'm not sure it's worth pursuing.


src/common/replayBrowser/dao.ts
```diff
@@ -0,0 +1,183 @@
+import { FileResult } from "common/replayBrowser";
+import { app } from "electron";
+import path from "path";
+import * as sqlite3 from "sqlite3";
+
+const db = new sqlite3.Database(path.join(app.getPath("userData"), "statsdb.sqlite"), (err) => {
```
vinceau, line 6: i think this is a better name
```suggestion
const db = new sqlite3.Database(path.join(app.getPath("userData"), "sqlippi.db"), (err) => {
```

src/common/replayBrowser/dao.ts
```diff
@@ -0,0 +1,183 @@
...
+       startTime     TEXT,
+       lastFrame     INTEGER,
+       playerCount   INTEGER,
+       player1       INTEGER,
+       player2       INTEGER,
+       player3       INTEGER,
+       player4       INTEGER,
+       settings      JSON,
+       metadata      JSON,
+       stats         JSON)
```
vinceau, line 25: Would kinda like to avoid having JSON as a column type. Would you consider having these as separate tables with typed/nullable information?

src/common/replayBrowser/dao.ts
```diff
@@ -0,0 +1,183 @@
...
+       fullPath      TEXT PRIMARY KEY,
+       name          TEXT,
+       folder        TEXT,
+       startTime     TEXT,
+       lastFrame     INTEGER,
+       playerCount   INTEGER,
+       player1       INTEGER,
+       player2       INTEGER,
+       player3       INTEGER,
+       player4       INTEGER,
```
vinceau, line 22: What are are these values for? It would be good if you could elaborate a bit on your database design decisions as either a reply or in the PR description.

vinceau: I would personally prefer having a table which only keeps track of file information (e.g. `fullPath`, `name`, `folder`), and then separates table for storing actual game information, keyed by full path, rather than everything in a single table.

vinceau (changes requested):
Overall pretty pleased with this PR. You'll need to rebase with the latest changes in `refactor/rewrite` where we introduced `threads.js` to handle worker threads. I'm allergic to plain JS so having the workers defined in JS should be avoided. Feel free to hit up the discord if you need help migrating the workers to use `threads.js`.

src/common/replayBrowser/ReplayClient.ts
```diff
@@ -0,0 +1,99 @@
+import { ipcRenderer } from "electron";
+
+import { FileLoadResult, FileResult, FolderResult } from "./types";
+
+export async function loadReplays(folder: string, progressCallback: (count: number, total: number) => void) {
```
vinceau, line 5: These event handlers should be refactored to use `electron-better-ipc`.

src/common/replayBrowser/ReplayServer.ts
```diff
@@ -0,0 +1,43 @@
+import { ipcMain } from "electron";
+
+import { getFullReplay, getPlayerReplays, pruneFolders } from "./dao";
```
vinceau, line 3: What does `dao` mean? I assume it's not "decentralized autonomous organization". Also on the topic of naming files, only React components should start with a capital letter. e.g. `SomeReactComponent.tsx` versus `someLibraryFunctions.ts`

package.json
```diff
@@ -13,14 +13,16 @@
...
+    "lint:fix": "yarn run lint --fix",
+    "postinstall": "install-app-deps"
   },
   "dependencies": {
     "@apollo/client": "^3.3.4",
     "@material-ui/core": "^4.11.2",
     "@material-ui/icons": "^4.11.2",
     "@material-ui/lab": "^4.0.0-alpha.57",
     "@slippi/slippi-js": "^5.1.1",
+    "@types/sqlite3": "^3.1.7",
```
vinceau, line 25: Types should go in `devDependencies`.

src/common/replayBrowser/ReplayClient.ts
alvarosevilla95: I checked out the library, it looks nice, but it seems oriented to request/response, not sure how to handle the callback messages to update the progress. Did you have something in mind?

src/common/replayBrowser/ReplayServer.ts
alvarosevilla95: Lol that's my Java showing, it stands for "Data Access Object". db.ts is probably a more suitable name
https://stackoverflow.com/questions/19154202/data-access-object-dao-in-java

alvarosevilla95: lovely :D

src/common/replayBrowser/dao.ts
alvarosevilla95: See my post on Discord. However looking at it, they should be TEXT fields!!! It's funny because ts was managing to convert them properly so all queries where working as expected 🤦

src/common/replayBrowser/ReplayClient.ts
alvarosevilla95: nvm I see you already have it on your branch

vinceau:
> thanks for the PR review. I've addressed most of the comments, but I'm not sure how to go about the db schema.
So I do agree a proper schema with all data unrolled and appropriate (de)normalization. However, with all the nested types in a game object, particularly the stats, we're looking at ~10 tables if properly unrolled, and the encoding and decoding is going to get a bit messy.
Now, on the other hand, given the current queries we're performing there's little benefit in having a proper schema anyway. The db size is minuscule (~20mb per 1k replays), and the js json decoding is fast enough to not really matter. I would however like to eventually move the core of global stats computing to actual sql queries and for that sort of stuff then the schema shine. There's also some questions about how/when data should be expected to be nullable or whether we should be using the player code or tag to identify players etc. that I would like some input on.
So anyway, for all that I would prefer to merge an initial version of the db with this simplified schema and then separately update the schema once I've developed the above a bit better. Performance is still way beyond acceptable from my experience. 
As to how the data is currently laid out, atm I'm just mimicking the FileResult interface, with all typed subfields (metadata,stats,settings) in json. Because we also want to easily query by player there's player{1-4} fields, currently their player codes (ideally we'd have both codes and tags). Then the file full path is used as primary key and the file's folder is also added so we can query/delete by it.

Responding to this here for better record keeping.

While I'm not opposed to having the tables be more fleshed out in another PR, since the tables aren't fully fleshed out, I would prefer to just have the table contain: 
```
       fullPath      TEXT PRIMARY KEY,
       name          TEXT,
       folder        TEXT,
       startTime     TEXT,
       lastFrame     INTEGER,
       settings      JSON,
       metadata      JSON,
```
And then we would just have to manually search through the metadata JSON for matching names and codes. As you mentioned the player information currently only contains player code, but it should contain both. I would much prefer either the SQL schema to be fully fleshed out or just not there for now, rather than being half-complete.

alvarosevilla95:
> > thanks for the PR review. I've addressed most of the comments, but I'm not sure how to go about the db schema.
> > So I do agree a proper schema with all data unrolled and appropriate (de)normalization. However, with all the nested types in a game object, particularly the stats, we're looking at ~10 tables if properly unrolled, and the encoding and decoding is going to get a bit messy.
> > Now, on the other hand, given the current queries we're performing there's little benefit in having a proper schema anyway. The db size is minuscule (~20mb per 1k replays), and the js json decoding is fast enough to not really matter. I would however like to eventually move the core of global stats computing to actual sql queries and for that sort of stuff then the schema shine. There's also some questions about how/when data should be expected to be nullable or whether we should be using the player code or tag to identify players etc. that I would like some input on.
> > So anyway, for all that I would prefer to merge an initial version of the db with this simplified schema and then separately update the schema once I've developed the above a bit better. Performance is still way beyond acceptable from my experience.
> > As to how the data is currently laid out, atm I'm just mimicking the FileResult interface, with all typed subfields (metadata,stats,settings) in json. Because we also want to easily query by player there's player{1-4} fields, currently their player codes (ideally we'd have both codes and tags). Then the file full path is used as primary key and the file's folder is also added so we can query/delete by it.
> 
> Responding to this here for better record keeping.
> 
> While I'm not opposed to having the tables be more fleshed out in another PR, since the tables aren't fully fleshed out, I would prefer to just have the table contain:
> 
> ```
>        fullPath      TEXT PRIMARY KEY,
>        name          TEXT,
>        folder        TEXT,
>        startTime     TEXT,
>        lastFrame     INTEGER,
>        settings      JSON,
>        metadata      JSON,
> ```
> 
> And then we would just have to manually search through the metadata JSON for matching names and codes. As you mentioned the player information currently only contains player code, but it should contain both. I would much prefer either the SQL schema to be fully fleshed out or just not there for now, rather than being half-complete.

Sounds good, the reason I was focused on the players is my other global stats PR which heavily relies on those queries, but tbf sqilte json querying should be good enough for now. 

So just to be clear, I'll have this table with the mentioned fields and an additional one with {fullPath,  stats} correct?

vinceau:
Yeah so i think something like this:

One table for keeping file system state:
```
       fullPath      TEXT PRIMARY KEY,
       filename          TEXT,
       folder        TEXT,
```

Another table for stats:
```
       fullPath      TEXT PRIMARY KEY,
       startTime     TEXT,
       lastFrame     INTEGER,
       settings      JSON,
       metadata      JSON,
```

src/main/replayBrowser/db.sql
```diff
@@ -0,0 +1,131 @@
...
+    playerIndex      INTEGER,
+    nametag          TEXT,
+    playerCode       TEXT,
+    port             INTEGER,
+    characterId      INTEGER,
+    characterColor   Integer,
+    startStocks      Integer,
+    endStocks        Integer,
+    playerType       Integer,
+    teamId           Integer,
```
vinceau, line 39: Why aren't these capitalized like the rest?

src/main/replayBrowser/db.sql
```diff
@@ -0,0 +1,131 @@
...
+    gameMode           INTEGER
+    FOREIGN KEY (gameId) REFERENCES replays(gameId) ON DELETE CASCADE
+);
+
+CREATE INDEX IF NOT EXISTS replay_data_start_time_path_idx ON replay_data(startTime)
+
+CREATE TABLE IF NOT EXISTS player (
+    gameId           INTEGER,
+    playerIndex      INTEGER,
+    nametag          TEXT,
```
vinceau, line 31: Can we separate `name` and `tag`? Where `tag` is like their slippi online tag and `name` is like their in-game name like from name-entry? I know you can't input name-entry online atm but we should still have that in the DB.

src/main/replayBrowser/db.sql
```diff
@@ -0,0 +1,131 @@
...
+    playerCode       TEXT,
+    port             INTEGER,
+    characterId      INTEGER,
+    characterColor   Integer,
+    startStocks      Integer,
```
vinceau, line 36: i feel like there isn't much point in storing `startStocks`? `endPercent` could be good for calculating tie-breaks when a game reaches time though.

vinceau (approved):
To avoid completely taking over the remote branch, I'll approve and merge this into `rewrite/db` and take over the database changes from there. Thanks for your contribution.

PR #127, opened 2021-02-26
Improve FileList rendering

alvarosevilla95:
Uses overscanCount to prerender rows and lazy caches stock and stage
imgs in memory.


alvarosevilla95:
[](url)

PR #134, opened 2021-03-28
Patchnote popup 

kreese89:
Uses localStorage to compare the most recent github release from the previous session with the most recent github release from the current session. If the two differ, opens a one-time modal that displays the most recent github release NewsItem. Clicking outside of the popup/pressing the escape key will close the modal. 

Popup renders after the NewsFeed finishes loading/rendering. Twitter feed will continue to load/render while the popup is open

Style can easily be changed if the current style is undesired.

Screen Shot 2021-03-28 at 7 01 22 PM


kreese89:
The popup renders when the NewsFeed finishes fetching the releases. Twitter feed continues to load in the background and will render while the popup is still open, which is nice

Issue #130, opened 2021-03-16
Dolphin Settings Page


NikhilNarayana:
The desktop app is now capable of launching dolphin for playback and netplay, but modifying dolphin requires going to the folder in appdata which isn't convenient. We have settings pages available which we can populate with various dolphin settings. There should be a page for Netplay Dolphin Settings and Playback Dolphin Settings, but most of the options will be the same.

Bolded items are required for this issue to be closed.
- [x] **Configure button**
- [x] **Reset button**
- [x] Clear Cache button
- [ ] Modify simple settings
  * ex: ability to turn on and off gecko codes

PR #140, opened 2021-05-25
Add gecko code configurator in DolphinSettings

RucksP:
### Type of change:
- [x] Feature

### Description:

Adds the ability to add and enable/disable Gecko Codes (part of Issue #130)

Rewrote [`Inifile`](https://github.com/dolphin-emu/dolphin/blob/master/Source/Core/Common/IniFile.cpp) from dolphin in typescript with some minor changes due to the lack of pass by reference in Javascript.
I haven't tested if the functions in `IniFile.ts` work if they are unrelated to adding/enabling gecko codes. There is 1 function which is not currently implemented.

_note_: the paths to the 2 dolphins are different for some reason so this only works for netplay dolphin atm.

### Screenshots:

addGecko

manageGecko


src/renderer/containers/Settings/DolphinSettings.tsx
```diff
@@ -87,12 +196,96 @@ export const DolphinSettings: React.FC<{ dolphinType: DolphinLaunchType }> = ({
...
+                      disableRipple
+                      onChange={handleToggle(i)}
+                    />
+                    <ListItemText primary={gecko.substring(1)} />
+                  </ListItem>
+                ))
+              )}
+            </List>
+            <Button fullWidth onClick={loadGeckoCodes} color="primary" variant="contained">
+              Refresh
```
NikhilNarayana, line 273: I'm not a huge fan of having a refresh button. The codes should always reload when the user hits save or changes the file being looked at. The only reason it is necessary in Dolphin is because sometimes dolphin can't tell that the editor is still open and will reload before the user has saved their changes.

NikhilNarayana (changes requested):
Do a rebase onto refactor/rewrite and drop the first three commits since those are no longer relevant.

## Major Notes
1. Codes and enabled info are also stored in the User folder. Changes should only ever be made to the User's settings since updates can and will overwrite the Sys codes as needed.
2. macOS data is saved inside the app bundle in `Slippi Dolphin.app/Contents/Resources/{Sys,User}/GameSettings`.
3. Linux is probably the hardest since the Appimage mounts a temp directory with a unique name each time for hosting all Sys files. The User files are generally in a consistent spot though (`~/.config/SlippiOnline` and `~/.config/SlippiPlayback`). I would be fine if we just disabled this feature for linux if it is too difficult.


### Side note
Reading this PR did make me realize that mainline Dolphin has much better handling for the gecko codes than I remember and since Slippi's has diverged I'll be updating it to be identical. The big change i noticed is the `Gecko_Disabled` section which fixes an issue we were having with codes before I added the `-` prefix for `Gecko_Enabled` to keep track of default codes that were disabled by the user.

RucksP:
Closing to make a create a new PR off of a much more up to date branch

PR #143, opened 2021-05-30
Rewrite settings to have a single source of truth

vinceau:
This PR moves all references of `electronSettings` to a `SettingsManager` class. Only this class should read/write to the `electronSettings` file.

With this PR we also slowly move towards relying less on node functions in the renderer thread. We start by moving the ISO file verification into the main thread requiring IPC to call the function. We also disable the Dolphin path verification for now since it's not entirely necessary at the moment and having more IPC calls to verify the Dolphin paths increase complexity.


NikhilNarayana (approved):
LGTM

PR #145, opened 2021-06-02
macOS-specific icons

ryanmcgrath:
These should be properly sized, tested across Dock and Finder.


vinceau:
Would you be able to replace all of them across the board? i.e. Windows and Linux too? Surely it won't look *bad*.

vinceau:
> 
> 
> Would you be able to replace all of them across the board? i.e. Windows and Linux too? Surely it won't look _bad_.

@ryanmcgrath Bump on this  ^. If it looks bad, we'll just revert it.

PR #149, opened 2021-06-14
Improve toast management

vinceau:
This PR adds `react-toast-notifications` for better toast management.

We can use the `useToasts()` hook for showing a new toast. We also use a custom toast container to stylize the toasts with `material-ui`. Additionally, we remove our old snackbar code from the old app state hook.


NikhilNarayana (approved):
LGTM

PR #153, opened 2021-06-19
Add mirroring, auto switcher, and discovery logic (basic UI)

NikhilNarayana:


vinceau (changes requested):

src/console/ipc.ts
```diff
@@ -0,0 +1,18 @@
+import { _, EmptyPayload, makeEndpoint, SuccessPayload } from "../ipc";
+import { DiscoveredConsoleInfo } from "./discovery";
+import { MirrorDetails } from "./types";
+
+export const addMirrorConfig = makeEndpoint.main("addMirrorConfig", <{ config: MirrorDetails }>_, <SuccessPayload>_);
```
vinceau, line 5: It's not good that `MirrorDetails` is accepted as the input parameters, since `MirrorDetails` is not entirely JSON serializable. We should only pass serializable JSON through IPC.

As a reminder, this is the type of `MirrorDetails`.

```ts
export interface MirrorDetails {
  id?: number;
  ipAddress: string;
  port: number;
  folderPath: string;
  isRealTimeMode: boolean;
  isRelaying?: boolean;
  connection?: ConsoleConnection | null;
  fileWriter?: SlpFileWriter | null;
  isMirroring?: boolean;
  obsSettings?: OBSSettings | null;
  obsManager?: OBSManager | null;
}
```

We should be more explicit in what values are allowed in the config, since we never expect the frontend to send things like `OBSManager` and `SlpFileWriter`s through IPC to the main.

src/console/main.ts
```diff
@@ -0,0 +1,27 @@
...
+
+addMirrorConfig.main!.handle(async ({ config }) => {
+  mirrorManager.start(config);
+  return { success: true };
+});
+
+startMirroring.main!.handle(async ({ ip }) => {
+  mirrorManager
+    .startMirroring(ip)
+    .catch((err: Error) => log.info(`[Mirroring] Failed to start mirroring Wii @ ${ip}\n${err}`));
```
vinceau, line 15: I don't think we should be suppressing this error with `.catch`. It's better if we just throw the error so we can alert the frontend when there's an issue.

```suggestion
  await mirrorManager.startMirroring(ip);
```

src/console/autoSwitcher.ts
```diff
@@ -0,0 +1,111 @@
...
+    this.obsPassword = settings.password;
+  }
+
+  private _getSceneSources = async () => {
+    // eslint-disable-line
+    const res = await this.obs.send("GetSceneList");
+    const scenes = res.scenes || [];
+    const pairs = _.flatMap(scenes, (scene) => {
+      const sources = scene.sources || [];
+      return _.map(sources, (source) => ({
```
vinceau, line 43:
```suggestion
      return sources.map((source) => ({
```

vinceau (changes requested):

src/console/discovery.ts
```diff
@@ -0,0 +1,151 @@
...
+import moment from "moment";
+
+import { discoverConsoleFound, discoverConsoleLost } from "./ipc";
+
+export interface DiscoveredConsoleInfo {
+  ip: string;
+  mac: string;
+  name: string | undefined;
+  timeout: NodeJS.Timeout;
+  firstFound: moment.Moment;
```
vinceau, line 12: Again, since `DiscoveredConsoleInfo` is sent to the renderer via IPC, the JSON we send should be serializable. `timeout` shouldn't be sent to the frontend, and `firstFound` should probably be an ISO string.

src/console/discovery.ts
```diff
@@ -0,0 +1,151 @@
...
+     *  u8 mac_addr[6]; // 10 - 16
+     *  unsigned char nickname[32]; // 16 - 48
+     */
+
+    // String for Wii's MAC address
+    const mac = [];
+    for (let i = 10; i <= 15; i += 1) {
+      mac.push(msg.readUInt8(i).toString(16).padStart(2, "0"));
+    }
+    const macAddr = `${mac[0]}:${mac[1]}:${mac[2]}:${mac[3]}:${mac[4]}:${mac[5]}`;
```
vinceau, line 49:
```suggestion
    const macAddr = mac.join(":");
```
Why not just do something like this?

src/console/autoSwitcher.ts
```diff
@@ -0,0 +1,111 @@
...
+      // if you send a password when authentication is disabled, OBS will still connect
+      await this.obs.connect(
+        {
+          address: this.obsIP,
+          password: this.obsPassword,
+        },
+        (err) => log.error(err?.message),
+      );
+      await this.obs.on("SceneItemAdded", async () => this._getSceneSources()); // eslint-disable-line
+      await this.obs.on("SceneItemRemoved", async () => this._getSceneSources()); // eslint-disable-line
```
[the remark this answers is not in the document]
NikhilNarayana: /shrug i don't remember and there are no complaints when i remove that comment on legacy

src/console/autoSwitcher.ts
```diff
@@ -0,0 +1,111 @@
+import log from "electron-log";
+import _ from "lodash";
+import OBSWebSocket from "obs-websocket-js";
+
+import { OBSSettings } from "./types";
+
+export class OBSManager {
```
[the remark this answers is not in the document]
NikhilNarayana: yeah i would agree that the `ObsManager` style looks the cleanest. I'll zip through real quick and refactor any similar names i can find.

src/console/main.ts
NikhilNarayana: this was leftover from when I was trying to figure out why mirroring wasn't working (it was cause I forgot the port). Removed it.

vinceau (approved):

PR #157, opened 2021-06-20
Add multiselect to replay browser

### LINKED ISSUE #155: Missing multiselect functionality in the rewrite
One of the features that hasn't been ported over in `refactor/rewrite` is the ability to multiselect replay files to play/delete. We should add back in the ability to multiselect files by clicking on them (i.e. without an explicit checkbox like in the current desktop app). A similar toolbar at the bottom with play all, clear, delete should show up when files are selected.

Eli-Caraballo:
Fixes #155 
Styles are temporary.
Functionality needs to be added still to "Play All" and "Delete" buttons

### DIFF
--- a/src/renderer/containers/ReplayBrowser/ReplayFile.tsx
+++ b/src/renderer/containers/ReplayBrowser/ReplayFile.tsx
@@ -28,8 +28,8 @@ export interface ReplayFileProps extends FileResult {
   onSelect: () => void;
   onPlay: () => void;
   onOpenMenu: (index: number, element: HTMLElement) => void;
-  handleAddToList: (name: string) => void;
-  list: Array<string>;
+  onClick: () => void;
+  selectedIndex: number;
 }
 
 export const ReplayFile: React.FC<ReplayFileProps> = ({
@@ -38,8 +38,8 @@ export const ReplayFile: React.FC<ReplayFileProps> = ({
   style,
   onSelect,
   onPlay,
-  handleAddToList,
-  list,
+  onClick,
+  selectedIndex,
   startTime,
   settings,
   name,
@@ -47,6 +47,7 @@ export const ReplayFile: React.FC<ReplayFileProps> = ({
   lastFrame,
   fullPath,
 }) => {
+  const selected = selectedIndex !== -1;
   const date = new Date(startTime ? Date.parse(startTime) : 0);
   let stageName = "Unknown";
   try {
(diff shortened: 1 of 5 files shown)


src/renderer/containers/ReplayBrowser/ReplayFile.tsx
```diff
@@ -28,6 +28,8 @@ export interface ReplayFileProps extends FileResult {
   onSelect: () => void;
   onPlay: () => void;
   onOpenMenu: (index: number, element: HTMLElement) => void;
+  handleAddToList: (name: string) => void;
+  list: Array<string>;
```
vinceau, line 32: I don't want `ReplayFile` to have to care about the list of items. I would rather it simply have a `selected` prop and an `onClick` prop that it gets passed from the parent.

```suggestion
  onClick: () => void;
  selected: boolean;
```

The basic styles you added should then only be shown if `selected` is true.

src/renderer/containers/ReplayBrowser/ReplayFile.tsx
```diff
@@ -83,19 +108,34 @@ export const ReplayFile: React.FC<ReplayFileProps> = ({
               <ReplayActionButton
                 label="More options"
                 onClick={(e) => {
+                  e.stopPropagation();
```
vinceau, line 111: Instead of passing in `e.stopPropagation()` manually in each of the `onClick` props, why not just set it to stop propagation at the `ReplayActionButton` level.

src/renderer/containers/ReplayBrowser/index.tsx
```diff
@@ -46,6 +46,7 @@ export const ReplayBrowser: React.FC = () => {
   const fileErrorCount = useReplays((store) => store.fileErrorCount);
   const rootSlpPath = useSettings((store) => store.settings.rootSlpPath);
   const { addToast } = useToasts();
+  const [list, setList] = React.useState<Array<string>>([]);
```
vinceau, line 49: Having the store here means that changing pages will clear the list, since the list gets reset on mount/unmount. You should instead modify `src/renderer/store/replays/index.ts` to add a list for selected items, and corresponding method for toggling the selection etc. That way the list state gets preserved even when changing pages.

src/renderer/containers/ReplayBrowser/index.tsx
```diff
@@ -161,17 +176,61 @@ export const ReplayBrowser: React.FC = () => {
...
+                        css={css`
+                          background-color: rgb(220, 219, 220);
+                          color: black;
+                          margin-left: 12px;
+                        `}
+                      >
+                        Delete
+                      </Button>
+                    </div>
+                  ) : null}
```
vinceau, line 228: Can you move this into a separate component? Call it like `FileSelectionToolbar.tsx` or something inside of `src/renderer/containers/ReplayBrowser`.

vinceau (changes requested):
Not bad for a first time contributor! Just a few things to fix up.

vinceau (approved):

PR #158, opened 2021-06-20
Add "Play from here" to kill/punish tables

NikhilNarayana:
image

### DIFF
--- a/src/dolphin/main.ts
+++ b/src/dolphin/main.ts
@@ -47,10 +47,11 @@ removePlayKeyFile.main!.handle(async () => {
   return { success: true };
 });
 
-viewSlpReplay.main!.handle(async ({ filePath }) => {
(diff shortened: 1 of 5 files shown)


src/dolphin/main.ts
```diff
@@ -47,10 +49,14 @@ removePlayKeyFile.main!.handle(async () => {
   return { success: true };
 });
 
-viewSlpReplay.main!.handle(async ({ filePath }) => {
+viewSlpReplay.main!.handle(async ({ filePath, startFrame }) => {
+  if (startFrame === undefined) {
+    startFrame = Frames.FIRST;
+  }
```
vinceau, line 55: I don't think we need this. `startFrame` can be left undefined according to the `ReplayCommunication` type as defined in `types`.ts.

vinceau (changes requested):

vinceau (approved):

PR #166, opened 2021-06-22
Clear Cache button and Confirmation Dialog

NikhilNarayana:


src/dolphin/manager.ts
```diff
@@ -108,6 +110,16 @@ export class DolphinManager extends EventEmitter {
...
     downloadAndInstallDolphin(launchType, log.info, true);
   }
+
+  public async clearCache(launchType: DolphinLaunchType) {
+    const userFolder = await findUserFolder(launchType);
+    const cacheFolder = path.join(userFolder, "Cache");
+    try {
+      fs.remove(cacheFolder);
+    } catch (err) {
+      log.error(err);
```
vinceau, line 120: Should we not throw a proper error here so the frontend can get the error?

src/renderer/components/ConfirmationModal.tsx
```diff
@@ -0,0 +1,55 @@
...
+    onSubmit();
+    onClose();
+  };
+
+  return (
+    <Dialog open={open} onClose={onClose} fullWidth={true} fullScreen={fullScreen} disableBackdropClick={true}>
+      <form onSubmit={handleSubmit}>
+        <StyledDialogTitle>{title}</StyledDialogTitle>
+        <DialogContent style={{ display: "flex" }}>
+          <div>{content}</div>
```
vinceau, line 35:
```suggestion
          <div>{content}</div>
          {children}
```
I think just remove the `content` prop and allow for children to be rendered.

vinceau (changes requested):

src/renderer/components/ConfirmationModal.tsx
```diff
@@ -0,0 +1,52 @@
...
+    console.log("submitting form...");
+    onSubmit();
+    onClose();
+  };
+
+  return (
+    <Dialog open={open} onClose={onClose} fullWidth={true} fullScreen={fullScreen} disableBackdropClick={true}>
+      <form onSubmit={handleSubmit}>
+        <StyledDialogTitle>{title}</StyledDialogTitle>
+        <DialogContent style={{ display: "flex" }}>{children}</DialogContent>
```
vinceau, line 33: Why is this `display: flex` here?

vinceau (changes requested):

src/renderer/components/ConfirmationModal.tsx
NikhilNarayana: I copied it from StartBroadcastDialog

vinceau (approved):

PR #182, opened 2021-06-24
feat: add back console relay

NikhilNarayana:
The style on the replay port needs to be fixed.

### DIFF
--- a/src/renderer/containers/Console/AddConnectionForm.tsx
+++ b/src/renderer/containers/Console/AddConnectionForm.tsx
@@ -90,8 +90,8 @@ export const AddConnectionForm: React.FC<AddConnectionFormProps> = ({ defaultVal
           <Toggle
             value={Boolean(isRealTimeMode)}
             onChange={(checked) => setValue("isRealTimeMode", checked)}
-            label="Real-time Mode"
-            description="When enabled, prevents delay from accumulating when mirroring. Keep this off unless both the Wii and computer are on a wired LAN connection."
+            label="Enable Real-time Mode"
+            description="Prevents delay from accumulating when mirroring. Keep this off unless both the Wii and computer are on a wired LAN connection."
           />
         </section>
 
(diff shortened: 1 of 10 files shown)


src/renderer/containers/Console/AddConnectionForm.tsx
```diff
@@ -108,50 +114,79 @@ export const AddConnectionForm: React.FC<AddConnectionFormProps> = ({ defaultVal
...
-              value={obsSourceName ?? ""}
-              required={showAdvanced}
-              onChange={(e) => setValue("obsSourceName", e.target.value)}
-            />
+              <Collapse in={showConnPort}>
+                <TextField
+                  label="Port Number"
+                  value={port ?? ""}
+                  required={showConnPort}
+                  onChange={(e) => setValue("port", Number(e.target.value.replace(/[^0-9]/g, "")))}
```
vinceau, line 186: This kinda replacement, where you convert numbers into text make the typing experience kinda wack cuz you cannot backspace to clear the field, it replaces it with zero. I would probably keep it as a string and maybe set it as a number on submit or something?

```suggestion
                  type="number"
                  onChange={(e) => setValue("port", e.target.value)}
```

vinceau (changes requested):

src/renderer/containers/Console/AddConnectionForm.tsx
```diff
@@ -108,50 +114,79 @@ export const AddConnectionForm: React.FC<AddConnectionFormProps> = ({ defaultVal
...
+                This connection's relay port will be shown on the console card after you have saved and is activated once you select connect."
               />
-              <TextField
-                label="OBS Password"
-                value={obsPassword ?? ""}
-                onChange={(e) => setValue("obsPassword", e.target.value)}
-                type="password"
+            </section>
+            <section>
+              <Toggle
```
vinceau, line 175: I don't really think this needs to be hidden behind a toggle. It should probably just a be titled "Connection Port" and have it be a text field which defaults to the Wii port.

vinceau (approved):

vinceau (approved):
LGTM

PR #185, opened 2021-06-25
feat: add functionality to change Display Name

NikhilNarayana:


src/renderer/components/ConfirmationModal.tsx
```diff
@@ -14,6 +14,9 @@ export interface ConfirmationModalProps {
   onSubmit: () => void;
   title: string;
   confirmText?: string;
+  cancelText?: string;
+  cancelColor?: "secondary" | "inherit" | "default" | "primary" | undefined;
```
vinceau, line 18: This doesn't need to be both optional and undefined.

```suggestion
  cancelColor?: "secondary" | "inherit" | "default" | "primary";
```

src/renderer/containers/Header/UserMenu.tsx
```diff
@@ -28,8 +25,9 @@ export const UserMenu: React.FC<{
   const refreshPlayKey = useAccount((store) => store.refreshPlayKey);
   const loading = useAccount((store) => store.loading);
   const [openLogoutPrompt, setOpenLogoutPrompt] = React.useState(false);
-  const theme = useTheme();
-  const fullScreen = useMediaQuery(theme.breakpoints.down("xs"));
+  const [name, setName] = React.useState(user.displayName);
```
vinceau, line 28: This shouldn't be a local React state. Since it means that other components that would want to render a user profile name would be out of date. `setDisplayName` should be a new method in the `useAccount` hook so that when updated all instances of the name get updated.

src/renderer/containers/Header/UserMenu.tsx
```diff
@@ -88,25 +108,38 @@ export const UserMenu: React.FC<{
...
+      >
+        <Typography paragraph={true}>Enter the new name in the field below</Typography>
+        <TextField
+          required={true}
+          id="displayName"
+          variant="outlined"
+          autoFocus={true}
+          key={`displayName`} // Change key on new mode to force field to reset
+          name="displayName"
+          value={nameChange || name}
```
vinceau, line 126: > `nameChange || name`
Doing this is not good. When you clear the entire input e.g. with Ctrl+backspace, it defaults to `name`. It makes for a really awkward typing experience. I would recommend creating a new component `NameChangeDialog` which has a local state for the input box value. You can also add a use effect hook to reset it to `user.displayName` before opening.

src/renderer/containers/Header/index.tsx
```diff
@@ -16,11 +16,11 @@ import React from "react";
 import { useToasts } from "react-toast-notifications";
 
 import { PlayIcon } from "@/components/PlayIcon";
+import { assertPlayKey } from "@/lib/gql";
```
vinceau, line 19: `gql.ts` is not a great name for a library. GraphQL is the method of fetching data but doesn't describe what the library actually does. I would prefer the file to be called `slippi.ts` or something because it interacts with the slippi backend. Whether it uses REST or GraphQL or websockets or IPC to do the changes doesn't matter so it shouldn't be named as such.

src/renderer/containers/Header/UserMenu.tsx
```diff
@@ -88,25 +108,38 @@ export const UserMenu: React.FC<{
...
-      <Dialog
-        fullScreen={fullScreen}
+      <ConfirmationModal
+        title="Change your Display Name"
+        open={openNameChangePrompt}
+        onClose={handleClose}
+        onSubmit={onChangeName}
+        fullWidth={false}
+      >
+        <Typography paragraph={true}>Enter the new name in the field below</Typography>
```
vinceau, line 118: I think this description is probably not needed.

```suggestion
```

src/renderer/containers/Header/UserMenu.tsx
```diff
@@ -88,25 +108,38 @@ export const UserMenu: React.FC<{
           Logout
         </MenuItem>
       </Menu>
-      <Dialog
-        fullScreen={fullScreen}
+      <ConfirmationModal
+        title="Change your Display Name"
```
vinceau, line 112:
```suggestion
        title="Change display name"
```

src/renderer/containers/Header/UserMenu.tsx
```diff
@@ -79,6 +89,16 @@ export const UserMenu: React.FC<{
...
           </MenuItem>
         )}
+        {playKey && (
+          <MenuItem
+            onClick={() => {
+              closeMenu();
+              setOpenNameChangePrompt(true);
+            }}
+          >
+            Edit Name
```
vinceau, line 99:
```suggestion
            Change display name
```

I think it might be better if it matched the modal title.

src/renderer/containers/Header/UserMenu.tsx
```diff
@@ -88,25 +108,38 @@ export const UserMenu: React.FC<{
...
+        <Typography paragraph={true}>Enter the new name in the field below</Typography>
+        <TextField
+          required={true}
+          id="displayName"
+          variant="outlined"
+          autoFocus={true}
+          key={`displayName`} // Change key on new mode to force field to reset
+          name="displayName"
+          value={nameChange || name}
+          label="Display Name"
```
vinceau, line 127:
```suggestion
          label="Display name"
```

src/renderer/containers/Header/UserMenu.tsx
```diff
@@ -88,25 +108,38 @@ export const UserMenu: React.FC<{
...
+        open={openNameChangePrompt}
+        onClose={handleClose}
+        onSubmit={onChangeName}
+        fullWidth={false}
+      >
+        <Typography paragraph={true}>Enter the new name in the field below</Typography>
+        <TextField
+          required={true}
+          id="displayName"
+          variant="outlined"
```
vinceau, line 122: I don't think we've used outlined inputs very much in the rest of the app forms. Maybe we should stick with "filled"?

```suggestion
          variant="filled"
```

vinceau (changes requested):

vinceau (approved):

vinceau (approved):

PR #186, opened 2021-06-26
Add help page

vinceau:


src/renderer/containers/Settings/HelpPage.tsx
```diff
@@ -0,0 +1,32 @@
...
+      <h1>Help</h1>
+      <MarkdownContent
+        content={faqMarkdown}
+        css={css`
+          max-width: 800px;
+          border-radius: 10px;
+          background-color: rgba(0, 0, 0, 0.5);
+          padding: 10px 20px;
+          * {
+            opacity: 0.7;
```
NikhilNarayana, line 20: the combination of background-color having half opacity and these having .7 makes a lot of text near impossible to read, we should up these values.

NikhilNarayana (changes requested):

src/renderer/containers/Settings/HelpPage.tsx
vinceau: Should be fixed in b6c702e

NikhilNarayana (approved):
****

PR #188, opened 2021-06-26
feat: ability to set play button functionality

### LINKED ISSUE #171: feat: add support for opening other ISOs easily
Right now people have to launch a configure dolphin or update their ISO to use UnclePunch Training Mode or 20XX. A way to launch dolphin from the home page without starting an ISO would suffice. A possibly more cohesive option would be to let the user add a `Training ISO` and have a way to start that specific ISO.

NikhilNarayana:
Fixes #171

image

### DIFF
--- a/src/renderer/containers/Settings/DolphinSettings.tsx
+++ b/src/renderer/containers/Settings/DolphinSettings.tsx
@@ -3,9 +3,6 @@ import { clearDolphinCache, configureDolphin, reinstallDolphin } from "@dolphin/
 import { DolphinLaunchType } from "@dolphin/types";
 import { css, jsx } from "@emotion/react";
 import Button from "@material-ui/core/Button";
-import FormControlLabel from "@material-ui/core/FormControlLabel";
-import Radio from "@material-ui/core/Radio";
-import RadioGroup from "@material-ui/core/RadioGroup";
 import { createStyles, makeStyles, Theme } from "@material-ui/core/styles";
 import Typography from "@material-ui/core/Typography";
 import { isLinux } from "common/constants";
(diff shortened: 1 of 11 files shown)


src/renderer/containers/Settings/DolphinSettings.tsx
```diff
@@ -42,12 +45,20 @@ const useStyles = makeStyles((theme: Theme) =>
 
 export const DolphinSettings: React.FC<{ dolphinType: DolphinLaunchType }> = ({ dolphinType }) => {
   const [dolphinPath, setDolphinPath] = useDolphinPath(dolphinType);
+  const [launchMeleeOnPlay, setLaunchMelee] = useLaunchMeleeOnPlay();
   const [modalOpen, setModalOpen] = React.useState(false);
   const classes = useStyles();
   const { addToast } = useToasts();
+
+  const onLaunchMeleeChange = async (event: any) => {
```
vinceau, line 53: Can you figure out the type of this event, to avoid using any?

src/renderer/containers/Settings/DolphinSettings.tsx
```diff
@@ -42,12 +45,20 @@ const useStyles = makeStyles((theme: Theme) =>
...
+    const value = event.target.value === "true" ? true : false;
```
vinceau, line 54:
```suggestion
    const value = event.target.value === "true";
```
Why not just this?

src/renderer/containers/Settings/DolphinSettings.tsx
```diff
@@ -80,6 +94,22 @@ export const DolphinSettings: React.FC<{ dolphinType: DolphinLaunchType }> = ({
...
+              control={<Radio />}
+            />
+            <FormControlLabel
+              value={false}
+              label="Launch Dolphin when clicking play on the home screen"
+              control={<Radio />}
+            />
+          </RadioGroup>
+        </SettingItem>
+      )}
```
vinceau, line 112: I don't really think this belongs here. I think it should one of the first options in the Game & Replays settings menu. I think I would prefer to keep the netplay and playback dolphin settings pretty much the same.

vinceau (changes requested):
This PR is pretty good for the most part. I do think the radio option descriptions could be more succinct. I might try think of better copy later.

vinceau (approved):

PR #190, opened 2021-06-26
feat: drag and drop multiple files

### LINKED ISSUE #174: feat: drag and drop multiple files
Drag and drop is a useful feature for sharing replays, but having to do them one at a time might not be ideal so there should be some way to drag and drop all the selected files on the replay browser.

FluentCoding:
This allows you to drag & drop multiple files when being in the selection mode in the replay browser.
To do this, simply drag the new added button "Drag & Drop Replays" to the place you want to drop it to. (maybe there is a better/cleaner/more user-friendly way but this was the first idea i came up with)

Successfully tested on Windows 10, worked perfectly with 3914 replay files being selected and dropped over in Windows Explorer.

Fixes #174
image

### DIFF
--- a/src/main/listeners.ts
+++ b/src/main/listeners.ts
@@ -13,9 +13,10 @@ import { fetchNewsFeedData } from "./newsFeed";
 import { verifyIso } from "./verifyIso";
 
 export function setupListeners() {
-  ipcMain.on("onDragStart", (event, filePath: string) => {
+  ipcMain.on("onDragStart", (event, filePaths: string[]) => {
+    console.log(filePaths);
     event.sender.startDrag({
-      file: filePath,
+      files: filePaths, // if this shows an error, its due to the development api being outdated
       icon: nativeImage.createFromPath(path.join(__static, "images", "file.png")),
     });
   });
--- a/src/main/listeners.ts
+++ b/src/main/listeners.ts
@@ -14,10 +14,12 @@ import { verifyIso } from "./verifyIso";
 
 export function setupListeners() {
   ipcMain.on("onDragStart", (event, filePaths: string[]) => {
-    event.sender.startDrag({
+    const dragArguments: any = {
       files: filePaths, // if this shows an error, its due to the development api being outdated
       icon: nativeImage.createFromPath(path.join(__static, "images", "file.png")),
-    });
+    };
+
+    event.sender.startDrag(dragArguments);
   });
 
   ipcMain.on("getAppSettingsSync", (event) => {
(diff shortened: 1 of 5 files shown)


src/main/listeners.ts
```diff
@@ -13,9 +13,9 @@ import { fetchNewsFeedData } from "./newsFeed";
 import { verifyIso } from "./verifyIso";
 
 export function setupListeners() {
-  ipcMain.on("onDragStart", (event, filePath: string) => {
+  ipcMain.on("onDragStart", (event, filePaths: string[]) => {
     event.sender.startDrag({
-      file: filePath,
+      files: filePaths, // if this shows an error, its due to the development api being outdated
```
NikhilNarayana, line 18:
```suggestion
      file: filePaths,
```

the API accepts `string[]` for `file`.

src/renderer/containers/ReplayBrowser/FileSelectionToolbar.tsx
```diff
@@ -53,13 +57,12 @@ export const FileSelectionToolbar: React.FC<FileSelectionToolbarProps> = ({
...
         <div>
-          <Button
-            color="secondary"
-            variant="contained"
-            size="small"
-            onClick={() => setShowDeletePrompt(true)}
-            startIcon={<DeleteIcon />}
-          >
+          <DraggableFiles fullPaths={filePaths}>
+            <Button color="secondary" variant="contained" size="small" startIcon={<DragIndicatorIcon />}>
```
NikhilNarayana, line 61:
```suggestion
            <Button
              color="secondary"
              variant="contained"
              size="small"
              startIcon={<DragIndicatorIcon />}
              css={css`
                cursor: inherit;
              `}
            >
```

Needs to inherit cursor otherwise you just get a pointer when hovering over the button.

src/renderer/containers/ReplayBrowser/FileSelectionToolbar.tsx
```diff
@@ -53,13 +57,12 @@ export const FileSelectionToolbar: React.FC<FileSelectionToolbarProps> = ({
...
+              Drag & Drop Replays
```
NikhilNarayana, line 62:
```suggestion
              Share
```

This is a bit long for the button and doesn't really tell the user what the usage it has. Maybe `Share` would work better.

NikhilNarayana (changes requested):

src/main/listeners.ts
FluentCoding: Yes, but I'm getting a runtime error when passing in a string[] into file, although the api recommends that file accepts `string | string[]`.

Using files instead worked out, but typescript marks an error there.

FluentCoding:
should be done now FYI

vinceau:
Superseded by #199

PR #191, opened 2021-06-26
deps: upgrade Electron to 13.1.8

NikhilNarayana:
This is a pretty big jump, but functionality doesn't require many fixes.

Main reasons to upgrade are:
* Launcher will run native on M1 and we still only ship a single app bundle.
* Revealing location of replays on Linux will select the file (tested on Ubuntu).

The only issue is that during dev we see a lot more `DevTools failed to load source map` warnings.

image

It would be ideal if we could fix this, but we haven't had much success. I'm making this PR in hopes that someone might poke their head in and have a solution.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -15,6 +15,13 @@ jobs:
       matrix:
         node-version: [14.x]
         os: [ubuntu-latest, windows-latest, macOS-latest]
+        include:
+          - os: windows-latest
+            package_command: yarn run package --publish never
+          - os: ubuntu-latest
+            package_command: yarn run package --publish never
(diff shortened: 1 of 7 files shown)


ryanmcgrath:
Is the source map thing the only bit blocking this?

NikhilNarayana:
> Is the source map thing the only bit blocking this?

It is a soft block, we might just deal with it even if it is annoying.

vinceau:
Does the sourcemap console log warnings get spammed via `electron-log`? Cuz if they do we should change that.

vinceau:
Closed in favour of #260

vinceau:
Closed in favour of #260

PR #192, opened 2021-06-27
refactor: add better ipc endpoint naming scheme

vinceau:
This PR renames the ipc endpoints to have a consistent naming scheme. All ipc endpoints now should have the prefix `ipc_`. Events from main -> renderer should also have the suffix `Event`. This clears up name space collisions and makes endpoints clearer.

I'm still not entirely convinced the interface for `makeEndpoint` needs to change. Cuz I think it is somewhat intuitive when you call `makEndpoint.main` you're making a main endpoint, and `makeEndpoint.renderer` you're making a renderer endpoint.


NikhilNarayana (approved):
I'm fine with this for now. I think we can make a better decision about changing makeEndpoint as more devs try it out.

PR #193, opened 2021-06-28
feat: dolphin migration if "Slippi Desktop App" folder in OS appData folder exists

NikhilNarayana:


src/renderer/containers/QuickStart/DolphinMigrationStep.tsx
```diff
@@ -0,0 +1,152 @@
+/** @jsx jsx */
+import { ipc_copyDolphin } from "@dolphin/ipc";
+import { DolphinLaunchType } from "@dolphin/types";
+import { css, jsx } from "@emotion/react";
+import styled from "@emotion/styled";
+import Box from "@material-ui/core/Box";
+import Container from "@material-ui/core/Container";
+import FormHelperText from "@material-ui/core/FormHelperText";
+import * as fs from "fs-extra";
```
vinceau, line 9: The renderer should not deal with node process things like filesystems. Anything that requires modifying files on the filesystem should dispatch a message to the main process. I think we might still be doing some `fs` stuff still in other renderer components but we really should aim not to.

src/renderer/containers/QuickStart/DolphinMigrationStep.tsx
```diff
@@ -0,0 +1,152 @@
...
+
+  const migrateNetplayDolphin = async () => {
+    await ipc_copyDolphin.renderer!.trigger({ dolphinType: DolphinLaunchType.NETPLAY, userPath: netplayPath });
+    setNetplayMigration(false);
+  };
+
+  const deleteOldDesktopAppFolder = async () => {
+    await fs.remove(oldDesktopAppPath);
+    setExists(false);
+  };
```
vinceau, line 37: Again this clean up stuff should be handled in the main process side, since websites (the renderer) shouldn't know anything about the filesystem.

src/renderer/containers/QuickStart/DolphinMigrationStep.tsx
```diff
@@ -0,0 +1,152 @@
...
+  } = useForm<{ netplayPath: string }>();
+  const netplayPath = watch("netplayPath");
+
+  const onNetplaySubmit = handleSubmit(migrateNetplayDolphin);
+
+  return (
+    <Box display="flex" flexDirection="column" flexGrow="1">
+      <Container>
+        <QuickStartHeader>Migrate Dolphin</QuickStartHeader>
+        {netplayMigration && (
```
vinceau, line 60: I feel like this "step" is really two steps since it handles both netplay migration and playback migration. I think I would prefer this to either be separate steps, with separate quick start components, or a single component where you can see both options all the time. e.g. check boxes for what you want to migrate: checkbox for netplay, and a checkbox for playback, where if netplay is checked a path input box shows That way it seems more like a single cohesive step.

vinceau (changes requested):

src/dolphin/main.ts
```diff
@@ -4,6 +4,7 @@ import {
   ipc_checkPlayKeyExists,
   ipc_clearDolphinCache,
   ipc_configureDolphin,
+  ipc_copyDolphin as ipc_copyDolphinConfig,
```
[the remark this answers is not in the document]
NikhilNarayana: this is VScode being dumb when renaming symbols

src/main/listeners.ts
```diff
@@ -43,4 +50,26 @@ export function setupListeners() {
...
+    });
+    return { success: true };
+  });
+
+  ipc_getFolderContents.main!.handle(async ({ path }) => {
+    const contents = await fs.readdir(path, { withFileTypes: true }).catch((err) => {
+      throw new Error(err);
+    });
+    return { success: true, contents };
+  });
```
vinceau, line 74: This kind of very general purpose things should be avoided, specially the `deleteFolder` call. IPC calls should be more specific to a single task. We can think of the front end as the client, which we shouldn't trust. If the frontend happens to load a rogue script, it could call `ipc_deleteFolder.renderer!.trigger({path: "C:\\Windows"})` for example. The IPC calls should be much more specific for example:
```ts
await ipc_handleMigration.renderer!.trigger({
  migrateNetplay: "C:\\Some\\Path",
  migratePlayback: false,
})
```

This way even if the frontend goes rogue, their scope is only limited to calling that one IPC function.

Obviously having `nodeIntegration: true` bypasses all of this security anyways, but it's still best practice to have specific IPC functions.

vinceau (changes requested):

src/common/ipc.ts
```diff
@@ -17,7 +17,11 @@ export const ipc_getDesktopAppPath = makeEndpoint.main(
   <{ path: string; exists: boolean }>_,
 );
 
-export const ipc_deleteFolder = makeEndpoint.main("deleteFolder", <{ path: string }>_, <SuccessPayload>_);
+export const ipc_migrateDolphin = makeEndpoint.main(
+  "deleteFolder",
```
vinceau, line 21: This event name should probably be updated.

```suggestion
  "migrateDolphin",
```

src/main/listeners.ts
```diff
@@ -59,10 +61,18 @@ export function setupListeners() {
...
+    if (migrateNetplay) {
+      await dolphinManager.copyDolphinConfig(DolphinLaunchType.NETPLAY, migrateNetplay);
+    }
+    if (migratePlayback) {
+      await dolphinManager.copyDolphinConfig(DolphinLaunchType.PLAYBACK, migratePlayback);
+    }
+    if (desktopAppPath) {
+      await fs.remove(desktopAppPath).catch((err) => {
+        throw new Error(err);
+      });
```
vinceau, line 74:
```suggestion
      await fs.remove(desktopAppPath);
```
There's no point catching then throwing the same error.

src/common/ipc.ts
```diff
@@ -17,7 +17,11 @@ export const ipc_getDesktopAppPath = makeEndpoint.main(
...
+  <{ migrateNetplay: string | null; migratePlayback: string | null; desktopAppPath: string | null }>_,
```
vinceau, line 22: Why is the old desktop app path passed in as a parameter? Don't we already know what the old desktop app path is?

vinceau (changes requested):

src/common/ipc.ts
NikhilNarayana: atm the main process just generates it and the renderer stores it. I could have the main process regenerate it or store it, but i'm not sure where would be a good place to store it.

vinceau: Why does it need to be stored? Why can't it just get returned to itself, the same way it gets sent to renderer?

vinceau (approved):

vinceau (approved):

PR #196, opened 2021-06-30
feat: add import button to DolphinSettings

NikhilNarayana:
Lets users import dolphin settings on macOS and Windows. Linux stores settings in a user config dir so we don't need to worry about them.

### DIFF
--- a/src/dolphin/ipc.ts
+++ b/src/dolphin/ipc.ts
@@ -35,7 +35,7 @@ export const ipc_launchNetplayDolphin = makeEndpoint.main("launchNetplayDolphin"
 
 export const ipc_migrateDolphin = makeEndpoint.main(
   "migrateDolphin",
-  <{ migrateNetplay: string | null; migratePlayback: boolean }>_,
+  <{ migrateNetplay: string | null; migratePlayback: boolean; migratePlaybackPath?: string }>_,
   <SuccessPayload>_,
 );
 
--- a/src/dolphin/ipc.ts
+++ b/src/dolphin/ipc.ts
@@ -33,7 +33,7 @@ export const ipc_viewSlpReplay = makeEndpoint.main("viewSlpReplay", <{ files: Re
 
 export const ipc_launchNetplayDolphin = makeEndpoint.main("launchNetplayDolphin", <EmptyPayload>_, <SuccessPayload>_);
 
-export const ipc_getDesktopAppDolphinInfo = makeEndpoint.main(
+export const ipc_checkDesktopAppDolphin = makeEndpoint.main(
   "getDesktopAppDolphinPath",
   <EmptyPayload>_,
   <{ dolphinPath: string; exists: boolean }>_,
--- a/src/dolphin/ipc.ts
+++ b/src/dolphin/ipc.ts
@@ -33,6 +33,12 @@ export const ipc_viewSlpReplay = makeEndpoint.main("viewSlpReplay", <{ files: Re
 
 export const ipc_launchNetplayDolphin = makeEndpoint.main("launchNetplayDolphin", <EmptyPayload>_, <SuccessPayload>_);
 
+export const ipc_getDesktopAppDolphinInfo = makeEndpoint.main(
+  "getDesktopAppDolphinPath",
+  <EmptyPayload>_,
+  <{ dolphinPath: string; exists: boolean }>_,
+);
+
 // toImportDolphin path must point to a "Slippi Dolphin.{exe,app}"
 export const ipc_importDolphinSettings = makeEndpoint.main(
   "importDolphinSettings",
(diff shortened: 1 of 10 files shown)


src/dolphin/ipc.ts
```diff
@@ -35,7 +35,7 @@ export const ipc_launchNetplayDolphin = makeEndpoint.main("launchNetplayDolphin"
 
 export const ipc_migrateDolphin = makeEndpoint.main(
   "migrateDolphin",
-  <{ migrateNetplay: string | null; migratePlayback: boolean }>_,
+  <{ migrateNetplay: string | null; migratePlayback: boolean; migratePlaybackPath?: string }>_,
   <SuccessPayload>_,
 );
```
vinceau, line 40: While I guess this is not the worst, do you think we can generalize this IPC endpoint over the two Dolphin types?

For example, maybe something like this:

1. `ipc_importDolphinSettings`: IPC endpoint which accepts a path string and a dolphin type (either netplay or dolphin) and just copies the folder over (and doesn't delete anything).

2. `ipc_completeDolphinMigrationStep`: IPC endpoint which accepts no parameters, but simply "completes the migration step" by deleting the old slippi desktop app folder.

I guess this does mean that for the migration step, the renderer will need to know the location of the old desktop app path, but for the purpose of the migration, not for the purpose of the completion (the folder deletion). The end of the migration quick start step, simply calls calls `complete` to remove the previous path.

What are your thoughts on something like this?

vinceau (changes requested):

src/renderer/containers/QuickStart/ImportDolphinSettingsStep.tsx
```diff
@@ -24,21 +27,32 @@ type FormValues = {
...
-    setExists(false);
+    if (values.shouldImportNetplay) {
+      await ipc_importDolphinSettings.renderer!.trigger({
+        toImportDolphinPath: values.netplayPath,
+        type: DolphinLaunchType.NETPLAY,
+      });
+    }
+    if (values.shouldImportPlayback) {
+      const binaryName = isMac ? "Slippi Dolphin.app" : "Slippi Dolphin.exe";
+      const playbackPath = path.join(desktopAppPath, "dolphin", binaryName);
```
vinceau, line 42: What are your thoughts on having `main` should just send `renderer` the path with the `Dolphin.exe` or `Dolphin.app` so `renderer` doesn't have to do this `path.join` stuff?

src/renderer/containers/Settings/DolphinSettings.tsx
```diff
@@ -1,14 +1,23 @@
...
 import Button from "@material-ui/core/Button";
 import CircularProgress from "@material-ui/core/CircularProgress";
+import FormHelperText from "@material-ui/core/FormHelperText";
 import { createStyles, makeStyles, Theme } from "@material-ui/core/styles";
 import Typography from "@material-ui/core/Typography";
-import { isLinux } from "common/constants";
+import { isLinux, isMac } from "common/constants";
 import { shell } from "electron";
+import log from "electron-log";
+import _ from "lodash";
```
vinceau, line 18:
```suggestion
import capitalize from "lodash/capitalize";
```

Can we just use the default capitalize import if it's the only function we're using from lodash?

vinceau (changes requested):

src/renderer/containers/QuickStart/ImportDolphinSettingsStep.tsx
NikhilNarayana: yeah makes sense, we can go that route

src/main/listeners.ts
```diff
@@ -58,13 +60,31 @@ export function setupListeners() {
...
     }
 
-    return { exists: exists };
+    const dolphinFolderPath = path.join(desktopAppPath, "dolphin");
+    exists = await fs.pathExists(dolphinFolderPath);
+
+    const dolphinExecutablePath = await findDolphinExecutable(DolphinLaunchType.NETPLAY, dolphinFolderPath);
+
+    return { dolphinPath: dolphinExecutablePath, exists: exists };
+  });
```
vinceau, line 81: Since this endpoint is now really only in charge of fetching and returning the old dolphin path, do you think we should move this endpoint into the `@dolphin/ipc` module instead?

vinceau (approved):
I'm pretty okay with this PR. Just not sure if we should move the dolphin path endpoint into the dolphin module. I'll let you decide.

src/main/listeners.ts
NikhilNarayana: yeah i wasn't sure about that part, also the name is kinda weird. It checks the existence of the desktop app folder and then the dolphin folder. I think i'll rename it and move it.

vinceau (approved):
LGTM

PR #198, opened 2021-07-03
feat: basic dolphin ini manipulation

NikhilNarayana:
* Port IniFile (grabbed from #140) and GeckoCodeConfig (renamed to just GeckoCode) to TS.
* Modify adding game path to use the new IniFile implementation.
* Add a way to set a `Boot to CSS` gecko code for people that want to play offline (need more discussion on when to call this).
  * This doesn't work for NTSC-J at the moment.

I want to merge this in and pawn the UI work for gecko code configuration off to anyone interested, with #140 author @RucksP given priority.


src/dolphin/util.ts
```diff
@@ -1,11 +1,13 @@
 import { settingsManager } from "@settings/settingsManager";
+import { app } from "electron";
 import log from "electron-log";
 import * as fs from "fs-extra";
-import * as ini from "ini";
```
vinceau, line 4: So if we previously used the `ini` package to parse the ini file, and now we're manually handling the ini parsing,  does that mean we don't need the `ini` package anymore?

NikhilNarayana: yep, i'll remove it.

src/dolphin/util.ts
```diff
@@ -77,20 +104,69 @@ export async function addGamePathToInis(gameDir: string): Promise<void> {
...
+  } else {
+    if (geckoCodes[bootCodeIdx].enabled === enable) {
+      return;
+    }
+    geckoCodes[bootCodeIdx].enabled = enable;
+  }
+
+  saveCodes(localIni, geckoCodes);
+
+  localIni.save(localIniPath);
```
vinceau, line 171: How come we load `globalIni` and `localIni` in the beginning of this function, but when we call save, we only save the localIni?

NikhilNarayana: we load inis from Sys and User to have complete knowledge of what codes are enabled and how. We only write to the ini in User because otherwise we would be overwriting the defaults. This is standard behavior in Dolphin.

src/dolphin/iniFile.ts
```diff
@@ -0,0 +1,278 @@
...
+        .replace(/['"]+/g, "");
+    }
+
+    return [keyOut, retValueOut];
+  }
+
+  /**Differs from IniFile.cpp by:
+   * returns section object, not pointer
+   */
+  public getSection(section_name: string): Section | undefined {
```
vinceau, line 44: I know this is ported from C++ but in this repo, we don't really use snake_case for variable names. So we should probably stick with conventions established here.

```suggestion
  public getSection(sectionName: string): Section | undefined {
```

src/dolphin/iniFile.ts
```diff
@@ -0,0 +1,278 @@
...
+/**
+ * The Section class
+ */
+export class Section {
+  public name: string;
+  public keys_order: string[];
+  public m_lines: string[];
+  public values: Map<string, string>;
+
+  public constructor(_name: string) {
```
vinceau, line 212:
```suggestion
  public constructor(name: string) {
```

Similar to the other suggestion, we probably wanna stick with the conventions laid out in this repo, where we haven't really been using underscore prefixes for params.

src/dolphin/iniFile.ts
```diff
@@ -0,0 +1,278 @@
...
+  }
+}
+
+/**
+ * The Section class
+ */
+export class Section {
+  public name: string;
+  public keys_order: string[];
+  public m_lines: string[];
```
vinceau, line 209: These should probably be camelCase too.

src/dolphin/iniFile.ts
```diff
@@ -0,0 +1,278 @@
...
+    }
+
+    return default_value;
+  }
+
+  public exists(key: string): boolean {
+    return this.values.get(key) !== undefined;
+  }
+
+  public delete = (key: string): boolean => {
```
vinceau, line 246:
```suggestion
  public delete(key: string): boolean {
```
We should probably be consistent with the method declarations.

src/dolphin/iniFile.ts
```diff
@@ -0,0 +1,278 @@
...
+
+  /**Differs from IniFile.cpp by:
+   * returns keys instead of passing it by reference
+   */
+  public getKeys(section_name: string): string[] | boolean {
+    const section = this.getSection(section_name);
+    if (section === undefined) {
+      return false;
+    }
+    return section.keys_order;
```
vinceau, line 95: Why does this return two different types? Why can't if section is undefined, it just return an empty array?

src/dolphin/iniFile.ts
```diff
@@ -0,0 +1,278 @@
...
+export class IniFile {
+  private sections: Section[];
+
+  public constructor() {
+    this.sections = [];
+  }
+
+  /** Differs from IniFile.cpp via:
+   * Instead of editing keyOut and valueOut by reference, return them */
+  public parseLine(line: string): Array<string | null> {
```
vinceau, line 21:
```suggestion
  public parseLine(line: string): [string, string] | [null, null] {
```
I don't think `Array<string | null>` is the correct type to return here.

src/dolphin/iniFile.ts
```diff
@@ -0,0 +1,278 @@
...
+    let retValueOut: string | null = null;
+    let keyOut: string | null = null;
+
+    if (line === "" || line[0] === "#") {
+      return [null, null];
```
vinceau, line 26:
```suggestion
      return [null, null] as const;
```

src/dolphin/iniFile.ts
```diff
@@ -0,0 +1,278 @@
...
+    const firstEquals = line.indexOf("=");
+    if (firstEquals !== -1) {
+      keyOut = line.substring(0, firstEquals).replace(/\s+/g, "");
+      retValueOut = line
+        .substring(firstEquals + 1)
+        .replace(/\s+/g, "")
+        .replace(/['"]+/g, "");
+    }
+
+    return [keyOut, retValueOut];
```
vinceau, line 38:
```suggestion
    return [keyOut, retValueOut] as const;
```

src/dolphin/iniFile.ts
```diff
@@ -0,0 +1,278 @@
...
+    return this.values.get(key) !== undefined;
+  }
+
+  public delete = (key: string): boolean => {
+    const success = this.values.delete(key);
+    if (!success) {
+      return false;
+    }
+    this.keys_order.splice(this.keys_order.indexOf(key), 1);
+    return true;
```
vinceau, line 252:
```suggestion
    if (success) {
      this.keys_order.splice(this.keys_order.indexOf(key), 1);
    }
    return success;
```

vinceau (changes requested):
There's a few typing issues that need to be fixed. As for the `snake_case` v `camelCase` I'll leave it up to you if you want to refactor them for code quality and consistency. I could understand why you might not want to but I think it might be worthwhile. Again, up to you.

src/dolphin/iniFile.ts
```diff
@@ -0,0 +1,278 @@
...
+
+  public async load(fileName: string, keep_current_data = true): Promise<boolean> {
+    if (!keep_current_data) {
+      this.sections = [];
+    }
+
+    const ins = fs.createReadStream(fileName);
+    ins.on("error", (e) => {
+      console.log("failed to read file with error", e);
+      alert(e);
```
vinceau, line 120:
```suggestion
      log.error("failed to read file:", e);
```

We really shouldn't be using `alert` for error handling.

src/dolphin/iniFile.ts
```diff
@@ -0,0 +1,278 @@
...
+
+    return true;
+  }
+
+  public save(filePath: string): boolean {
+    const out = fs.createWriteStream(filePath);
+
+    out.on("error", (e) => {
+      console.log("failed to write file with error", e);
+      alert(e);
```
vinceau, line 174:
```suggestion
      log.error("failed to write file", e);
```

Again, we probably shouldn't use `alert`.

src/dolphin/geckoCode.ts
```diff
@@ -0,0 +1,141 @@
...
+    if (ini === globalIni) {
+      gcodes.forEach((gcode) => {
+        gcode.defaultEnabled = gcode.enabled;
+      });
+    }
+  });
+  return gcodes;
+}
+
+function _readEnabledOrDisabled(iniFile: IniFile, section: string, enabled: boolean, codes: GeckoCode[]) {
```
vinceau, line 83: Can we refactor these to not have the `_` prefix too? They aren't private methods so it's fine to leave them without the prefix.

src/dolphin/iniFile.ts
```diff
@@ -0,0 +1,277 @@
...
+    if (section === undefined) {
+      return false;
+    }
+    return section.delete(key);
+  }
+
+  /**Differs from IniFile.cpp by:
+   * returns keys instead of passing it by reference
+   */
+  public getKeys(section_name: string): string[] {
```
vinceau, line 91:
```suggestion
  public getKeys(sectionName: string): string[] {
```

There's still a lot of snake case so maybe it's best to just find and replace them all.

src/dolphin/iniFile.ts
```diff
@@ -0,0 +1,277 @@
...
+    if (section === undefined) {
+      return [];
+    }
+    return section.keysOrder;
+  }
+
+  /**Differs from IniFile.cpp by:
+   * returns lines instead of passing it by reference
+   */
+  public getLines(section_name: string, remove_comments = false): string[] {
```
vinceau, line 102:
```suggestion
  public getLines(sectionName: string, removeComments = false): string[] {
```

src/dolphin/iniFile.ts
```diff
@@ -0,0 +1,277 @@
...
+    if (section === undefined) {
+      return [];
+    }
+
+    const lines = section.getLines(remove_comments);
+
+    return lines;
+  }
+
+  public async load(fileName: string, keep_current_data = true): Promise<boolean> {
```
vinceau, line 113:
```suggestion
  public async load(fileName: string, keepCurrentData = true): Promise<boolean> {
```

vinceau (changes requested):

vinceau (dismissed):

PR #199, opened 2021-07-04
Feat/multi drag drop

cnkeats:
This is an expansion on #190 that removes the "Share" button and instead allows users to drag the replay entry while also preserving the ability to drag the filename if no replays are selected.


FluentCoding:
I feel like this is a really good idea!

vinceau (approved):

vinceau (approved):

Issue #177, opened 2021-06-23
Feat: Gecko Code enabling and disabling


NikhilNarayana:
Some work has already been done in #140 and https://github.com/project-slippi/slippi-launcher/tree/rewrite/basic-ini-manip.

We want the ability to modify gecko codes in the dolphin settings to make it more accessible for users.

PR #201, opened 2021-07-06
feat: gecko code configurator

RucksP:
### Type of change:
- [x] Feature

### Description:

Adds the ability to add/remove and enable/disable Gecko Codes (Issue #177)
Created UI to add/toggle gecko codes using GeckoCode.ts and IniFile.ts

### Images

Dolphin Settings

Add Geckos

Toggle Geckos

### DIFF
--- a/src/dolphin/geckoCode.ts
+++ b/src/dolphin/geckoCode.ts
@@ -124,7 +124,7 @@ function makeGeckoCode(code: GeckoCode, lines: string[]) {
 
   lines.push(makeGeckoCodeTitle(code));
 
-  code.notes.forEach((line) => lines.push(`* ${line}`));
+  code.notes.forEach((line) => lines.push(`*${line}`));
 
   code.codeLines.forEach((line) => lines.push(line));
 }
@@ -197,12 +197,12 @@ export function removeGeckoCode(geckoCodeName: string, codes: GeckoCode[]) {
 
 export function geckoCodeToRaw(code: GeckoCode) {
   let rawGecko = makeGeckoCodeTitle(code);
-  code.notes.forEach((line) => (rawGecko = rawGecko.concat("\n", `* ${line}`)));
+  code.notes.forEach((line) => (rawGecko = rawGecko.concat("\n", `*${line}`)));
   code.codeLines.forEach((line) => (rawGecko = rawGecko.concat("\n", line)));
   return rawGecko;
 }
 
-export function setEnabledDisabled(gCodes: GeckoCode[], tCodes: TruncGeckoCode[]) {
+export function setEnabledDisabledFromTCodes(gCodes: GeckoCode[], tCodes: TruncGeckoCode[]) {
   tCodes.forEach((tCode) => {
     const gCodeIndex = gCodes.findIndex((gCode) => gCode.name === tCode.name);
     if (gCodeIndex !== -1) {
--- a/src/dolphin/geckoCode.ts
+++ b/src/dolphin/geckoCode.ts
@@ -20,7 +20,7 @@ export function loadGeckoCodes(globalIni: IniFile, localIni?: IniFile): GeckoCod
     const lines: string[] = ini.getLines("Gecko", false).filter((line) => {
       return line.length !== 0 || line[0] !== "#";
     });
-    let gcode: GeckoCode = {
+    const gcode: GeckoCode = {
       name: "",
       creator: "",
       enabled: false,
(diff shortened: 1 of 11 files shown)


src/dolphin/main.ts
```diff
@@ -5,23 +5,111 @@ import path from "path";
...
+  findUserFolder,
+  getUserIni,
+  loadCodes,
+  updateBootToCssCode,
+} from "./util";
+
+ipc_fetchGeckoCodes.main!.handle(async ({ dolphinType, iniName }) => {
+  console.log("fetching gecko codes...");
+  const gCodes = await loadCodes(dolphinType, iniName);
+  const tCodes: GeckoCode[] = [];
```
NikhilNarayana, line 44: what is the point of this? gCodes is already an array of gecko codes that you could just return. Also why is it called tCodes?

src/dolphin/geckoCode.ts
```diff
@@ -135,7 +135,72 @@ export function saveCodes(iniFile: IniFile, codes: GeckoCode[]) {
...
+    creator: "",
+    enabled: true,
+    defaultEnabled: false,
+    userDefined: true,
+    notes: [],
+    codeLines: [],
+  };
+
+  //fill out gecko info
+  rawGeckoLines.forEach((line) => {
```
NikhilNarayana, line 156: this switch case section is just reused from loadCodes so we should just make a new function to avoid duplication.

src/dolphin/main.ts
RucksP: tCodes is for truncated gecko codes, the notes and codeLines are empty in tCodes as opposed to gCodes so we don't end up storing codelines/notes (which can be very large) in the state of dolphinSettings

NikhilNarayana: gotcha, adding that as a comment would be a good idea.

we can also use the spread operator like so to simplify the code a bit
```ts
const tCode: GeckoCode = {
      ...gcode
      notes: [],
      codeLines: [],
    };
```
or just modify gCode in place and push that into tCodes. Either option is a bit cleaner.

src/renderer/containers/Settings/DolphinSettings.tsx
```diff
@@ -1,15 +1,36 @@
...
+  ipc_deleteGecko,
+  ipc_fetchGeckoCodes,
+  ipc_fetchSysInis,
   ipc_importDolphinSettings,
   ipc_reinstallDolphin,
+  ipc_toggleGeckos,
 } from "@dolphin/ipc";
 import { DolphinLaunchType } from "@dolphin/types";
 import { css, jsx } from "@emotion/react";
+import { Box, Tab, Tabs } from "@material-ui/core";
```
NikhilNarayana, line 17: You can't import components from material-ui this way without breaking things. Update the imports to be like the ones below.

src/renderer/containers/Settings/DolphinSettings.tsx
```diff
@@ -24,8 +45,23 @@ import { useDolphinPath } from "@/lib/hooks/useSettings";
...
 import { SettingItem } from "./SettingItem";
 
+function TabPanel(props: any) {
+  const { children, value, index, ...other } = props;
+
+  return (
+    <div role="tabpanel" hidden={value !== index} id={`full-width-tabpanel-${index}`} {...other}>
+      {value === index && (
+        <Box p={3}>
+          <Typography>{children}</Typography>
```
NikhilNarayana, line 55: This typography seems to be giving a bunch of errors in dev tools since it is equivalent to a p element and you can't put some elements inside that like forms.

src/renderer/containers/Settings/DolphinSettings.tsx
```diff
@@ -168,3 +205,222 @@ const ImportDolphinConfigForm: React.FC<{
...
+                disableRipple
+                onChange={() => {
+                  gecko.enabled = !gecko.enabled;
+                  setGeckoCodes([...geckoCodes]);
+                }}
+              />
+              <ListItemText primary={gecko.name} />
+              <IconButton>
+                <AssignmentIcon
+                  onClick={async () => {
```
NikhilNarayana, line 278: the onClick handler should be on the IconButton component.

src/renderer/containers/Settings/DolphinSettings.tsx
```diff
@@ -168,3 +205,222 @@ const ImportDolphinConfigForm: React.FC<{
...
+                    await navigator.clipboard.writeText(rawGecko !== undefined ? rawGecko : "");
+                    addToast(`Copied "${gecko.name}" To Clipboard`, { appearance: "success", autoDismiss: true });
+                  }}
+                />
+              </IconButton>
+              {gecko.userDefined ? (
+                <IconButton>
+                  <DeleteIcon
+                    onClick={async () => {
+                      if (window.confirm(`Are You Sure You Want To Delete "${gecko.name}"?`)) {
```
NikhilNarayana, line 289: Don't use a native window confirm, use the confirmation dialog component.

src/renderer/containers/Settings/DolphinSettings.tsx
```diff
@@ -168,3 +205,222 @@ const ImportDolphinConfigForm: React.FC<{
...
+                        await deleteGeckoHandler(gecko.name);
+                      }
+                    }}
+                  />
+                </IconButton>
+              ) : (
+                <LockIcon />
```
NikhilNarayana, line 296: Put inside an IconButton or it won't look right with user added codes.

src/renderer/containers/Settings/DolphinSettings.tsx
```diff
@@ -168,3 +205,222 @@ const ImportDolphinConfigForm: React.FC<{
...
+                label="Paste Gecko Code Here"
+                variant="outlined"
+                margin="normal"
+                rows="18"
+                onChange={({ target: { value } }) => setNewGeckoCodeRaw(value)}
+                multiline
+                fullWidth
+                required
+              ></TextField>
+              <Button type="submit" fullWidth variant="outlined" color="secondary">
```
NikhilNarayana, line 411: Use contained buttons instead of outlined.

src/renderer/styles/_fonts.scss
```diff
@@ -8,3 +8,6 @@
 @import "~@fontsource/maven-pro/500.css";
 @import "~@fontsource/maven-pro/600.css";
 @import "~@fontsource/maven-pro/700.css";
+
+//Google
+@import url("https://fonts.googleapis.com/css2?family=Space+Mono&display=swap");
```
NikhilNarayana, line 13: we can use fontsource (`yarn add @fontsource/space-mono`)

PR #203, opened 2021-07-09
feat: multi drag drop - file count on drag icon

cnkeats:
This is an addition to #199 that adds an indicator of how many files are selected while dragging, even when dragging out of the application window.

The process for dynamically updating the count is to first update the div that will be used for the icon. The div can be located anywhere as long as it is not visible to the user (currently it is in ReplayBrowser.tsx). The div is then converted to a dataURL which then gets passed over with the file list and then used as a nativeImage for the drag icon.

### DIFF
--- a/package.json
+++ b/package.json
@@ -38,6 +38,7 @@
     "firebase": "^8.1.2",
     "fs-extra": "^9.0.1",
     "graphql": "^15.4.0",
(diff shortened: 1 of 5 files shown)


vinceau:
I appreciate this attempt but the conversion from div -> image url is very slow. I don't think it's really worth that performance hit to just show the number.

PR #204, opened 2021-07-11
feat: multiple root replay directories

### LINKED ISSUE #131: feat: multiple root replay directories
Some people might want to store replays in multiple folders across multiple drives.

- [ ] Netplay Replay Folder must always exist and be a valid path.
- [ ] Ability to add and delete additional replay folders
- [ ] Replays folders must show up in the replays tab

cnkeats:
Fixes #131

Includes duplication checking to prevent users from re-selecting a folder already selected as root or as an extra
Includes duplication checking if a newly selected folder is a child of an already selected one

If checking for being a child will be removed, be aware that there is some weirdness when the folder exists multiple times in the replay browser and is selected.


cnkeats:
Updated with new settings design
image

NikhilNarayana (approved):
LGTM

PR #207, opened 2021-07-16
refactor: shift select respects direction

DomHynes:
This PR checks the indexes when shift selecting replays, then flips the array if the user has shift selected a newer replay after an older one.

This means it's possible to queue up vods in chronological order with shift select.


DomHynes:
I just realised the problem this solves (chronological shift selecting) can be worked around by sorting by Least Recent, but this still updates the shift select behaviour to be more idiomatic.

vinceau (approved):
LGTM

PR #208, opened 2021-07-25
feat: move broadcast and spectate to node worker

NikhilNarayana:


src/broadcast/main.ts
```diff
@@ -1,27 +1,29 @@
...
-  await spectateManager.refreshBroadcastList();
+  const sWorker = await spectateWorker;
+  await sWorker.refreshBroadcastList(authToken);
   return { success: true };
 });
 
 ipc_watchBroadcast.main!.handle(async ({ broadcasterId }) => {
+  const sWorker = await spectateWorker;
   const folderPath = settingsManager.get().settings.spectateSlpPath;
-  spectateManager.watchBroadcast(broadcasterId, folderPath, true);
```
vinceau, line 15: We used to have singleton to be `true` when calling `spectateManager.watchBroadcast()`. This doesn't seem to be the case anymore. Is this intentional?

src/broadcast/spectateWorker.ts
```diff
@@ -0,0 +1,82 @@
...
+  logSubject.next(msg);
+});
+
+spectateManager.on(SpectateEvent.ERROR, async (errorMsg: string) => {
+  errorSubject.next(errorMsg);
+});
+
+spectateManager.on(SpectateEvent.PLAY_FILE, async (playbackId: string, replayComm: ReplayCommunication) => {
+  spectateDetailsSubject.next({ playbackId, replayComm });
+});
```
vinceau, line 49: I'm not really a fan of spectate worker emitting a "play file" event. It shouldn't really need to know about playing things back. I think that should be main's job. The event should probably be called `NEW_GAME` or `NEW_FILE` or `GAME_START` or something which should return an Observable string (the SLP filename). then it should be main's job to generate the playback ID and replay communication.

src/broadcast/types.ts
```diff
@@ -14,3 +16,46 @@ export interface StartBroadcastConfig {
   authToken: string;
   name?: string;
 }
+
+export enum BroadcastEvent {
+  slippiStatusChange = "slippiStatusChange",
+  dolphinStatusChange = "dolphinStatusChange",
+  error = "broadcastError",
+  LOG = "log",
```
vinceau, line 24: The log event doesn't match the naming convention of the other events. We should probably coerce the others to be in upper snake case or leave it for another PR.

src/broadcast/workerInterfaces.ts
```diff
@@ -0,0 +1,114 @@
...
+const broadcastLog = electronLog.scope("broadcastManager");
+const spectateLog = electronLog.scope("spectateManager");
+
+export const broadcastWorker: Promise<Thread & BroadcastWorkerMethods> = new Promise((resolve, reject) => {
+  log.debug("broadcast: Spawning worker");
+
+  spawn<BroadcastWorkerSpec>(new Worker("./broadcastWorker"), { timeout: 30000 })
+    .then((worker) => {
+      worker.getDolphinStatusObservable().subscribe(({ status }) => {
+        log.info(`got dolphin status: ${status}`);
```
vinceau, line 26: Should we remove these logs? I think they were useful for debugging but they're probably not useful now.

vinceau (changes requested):

src/broadcast/workerInterfaces.ts
NikhilNarayana: yeah we can clean those up

src/broadcast/types.ts
NikhilNarayana: whoops, didn't notice the difference when I copied it from SpectateEvents.

src/broadcast/main.ts
NikhilNarayana:
```ts
  public watchBroadcast(broadcastId: string, targetPath: string, singleton?: true)
  ```
 the function signature forces true so I skipped it when i was rewriting. Do we want to put true anyways?

File: src/broadcast/spectateWorker.ts (hunk 2 above)
Comment (NikhilNarayana) [reply src/broadcast/spectateWorker.ts:49]: Sure

File: src/broadcast/main.ts (hunk 1 above)
Comment (vinceau) [reply src/broadcast/main.ts:15]: I didn't realise it was true by default. It's fine to leave it as undefined.

File: src/broadcast/broadcastManager.ts [hunk 5]
```diff
@@ -83,20 +82,20 @@ export class BroadcastManager {
       try {
         await this._connectToDolphin(config.ip, config.port);
       } catch (err) {
-        log.warn(err);
+        this.emit(BroadcastEvent.LOG, err);
```
vinceau, line 85: Error is an object. I'm pretty sure subscribers will be expecting a string. So we should make sure we're only return a string.

src/broadcast/broadcastManager.ts
```diff
@@ -70,7 +69,7 @@ export class BroadcastManager {
     });
     this.dolphinConnection.on(ConnectionEvent.ERROR, (err) => {
       // Log the error messages we get from Dolphin
-      log.error("[Broadcast] Dolphin connection error\n", err);
+      this.emit(BroadcastEvent.LOG, "[Broadcast] Dolphin connection error\n", err);
```
vinceau, line 72: Again over here, this event should only emit a single string.  We're returning 2 things here, a string, and an error object.

src/broadcast/types.ts
```diff
@@ -14,3 +16,46 @@ export interface StartBroadcastConfig {
   authToken: string;
   name?: string;
 }
+
+export enum BroadcastEvent {
+  SLIPPI_STATUS_CHANGE = "slippiStatusChange",
+  DOLPHIN_STATUS_CHANGE = "dolphinStatusChange",
+  ERROR = "broadcastError",
+  LOG = "log",
```
vinceau, line 24:
```suggestion
  SLIPPI_STATUS_CHANGE = "SLIPPI_STATUS_CHANGE",
  DOLPHIN_STATUS_CHANGE = "DOLPHIN_STATUS_CHANGE",
  ERROR = "ERROR",
  LOG = "LOG",
```

While we're refactoring these, can we make it like this?

src/broadcast/types.ts
```diff
@@ -14,3 +16,46 @@ export interface StartBroadcastConfig {
...
+}
+
+export enum SpectateEvent {
+  ERROR = "error",
+  BROADCAST_LIST_UPDATE = "broadcastListUpdate",
+  NEW_FILE = "newFile",
+  LOG = "log",
```
vinceau, line 31:
```suggestion
  ERROR = "ERROR",
  BROADCAST_LIST_UPDATE = "BROADCAST_LIST_UPDATE",
  NEW_FILE = "NEW_FILE",
  LOG = "LOG",
```

vinceau (changes requested):
Just some minor fixes.

src/broadcast/broadcastManager.ts
vinceau: We can probably log `err.message || JSON.stringify(err)`.

src/broadcast/spectateManager.ts
```diff
@@ -131,23 +98,25 @@ export class SpectateManager extends EventEmitter {
     await new Promise<void>((resolve, reject) => {
       const socket = new WebSocketClient();
 
-      socket.on("connectFailed", (error) => {
-        log.error(`[Spectate] WS connection failed\n`, error);
-        this.emit(SpectateManagerEvent.ERROR, error);
+      socket.on("connectFailed", (err) => {
+        const errMsg = err.message || JSON.stringify(err);
+        this.emit(SpectateEvent.LOG, `WS connection failed\n${errMsg}`);
+        this.emit(SpectateEvent.ERROR, errMsg);
```
vinceau, line 104: Can we use the `log` event for logs and the `error` event for errors? It seems like in a lot of places we're doing the double emit to log and error the same message.

vinceau (approved):
LGTM

PR #211, opened 2021-08-02
feat: integrated online enabler

NikhilNarayana:
UI needs some work, but the functionality is there.


cnkeats (dismissed):
The paid option to change connect code could be added to the user menu too with a link to the profile page on slippi.gg until that functionality is added to the launcher.

LGTM otherwise.

vinceau (approved):

PR #212, opened 2021-08-08
feat: update dolphin settings

cnkeats:
Pushes settings from Launcher to Dolphin when Dolphin is launched. Launcher settings are preferred and used to overwrite Dolphin settings when reasonable (time to launch dolphin is not noticeably affected).
- Replay root directory
- Use monthly subfolders

Users that don't use the Launcher will be able to change their settings as normal. However, if everyone is eventually required to use the Launcher, the ability to change settings via Dolphin should either be removed or synced with the Launcher.

image

### DIFF
--- a/src/dolphin/manager.ts
+++ b/src/dolphin/manager.ts
@@ -9,6 +9,7 @@ import { downloadAndInstallDolphin } from "./downloadDolphin";
 import { DolphinInstance, PlaybackDolphinInstance } from "./instance";
 import { DolphinLaunchType, ReplayCommunication } from "./types";
 import { addGamePathToIni, findDolphinExecutable, findUserFolder } from "./util";
+import { updateDolphinReplayPath } from "@dolphin/util";
 
 const log = electronLog.scope("dolphin/manager");
 
@@ -46,6 +47,8 @@ export class DolphinManager extends EventEmitter {
       throw new Error("Netplay dolphin is already open!");
     }
 
+    this.syncDolphinPath();
+
     const dolphinPath = await findDolphinExecutable(DolphinLaunchType.NETPLAY);
     log.info(`Launching dolphin at path: ${dolphinPath}`);
     const launchMeleeOnPlay = settingsManager.get().settings.launchMeleeOnPlay;
@@ -62,6 +65,9 @@ export class DolphinManager extends EventEmitter {
 
   public async configureDolphin(launchType: DolphinLaunchType) {
     log.debug(`configuring ${launchType} dolphin...`);
+
+    this.syncDolphinPath();
+
     const dolphinPath = await findDolphinExecutable(launchType);
     if (launchType === DolphinLaunchType.NETPLAY && !this.netplayDolphinInstance) {
       const instance = new DolphinInstance(dolphinPath);
@@ -125,6 +131,12 @@ export class DolphinManager extends EventEmitter {
       const gameDir = path.dirname(isoPath);
       await addGamePathToIni(launchType, gameDir);
     }
+  }
+
+  private async syncDolphinPath() {
+    // It's possible for the Launcher and Dolphin's replay direcories to become
+    // desynced if the user sets the directory via Dolphin, so update Dolphin's to match the launcher
+    updateDolphinReplayPath(settingsManager.getRootSlpPath());
   }
 
   public async clearCache(launchType: DolphinLaunchType) {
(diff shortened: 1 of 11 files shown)


src/dolphin/manager.ts
```diff
@@ -46,6 +46,8 @@ export class DolphinManager extends EventEmitter {
       throw new Error("Netplay dolphin is already open!");
     }
 
+    void updateDolphinSettings();
```
vinceau, line 49: Why are we `void`ing these promises?

```suggestion
    await updateDolphinSettings();
```

src/dolphin/manager.ts
```diff
@@ -62,6 +64,9 @@ export class DolphinManager extends EventEmitter {
 
   public async configureDolphin(launchType: DolphinLaunchType) {
     log.debug(`configuring ${launchType} dolphin...`);
+
+    void updateDolphinSettings();
```
vinceau, line 68:
```suggestion
    await updateDolphinSettings();
```

src/renderer/components/PathInput.tsx
```diff
@@ -30,9 +32,13 @@ export const PathInput = React.forwardRef<HTMLInputElement, PathInputProps>((pro
         <CustomInput inputRef={ref} disabled={true} value={value} placeholder={placeholder} />
         {endAdornment}
       </InputContainer>
-      <Button color="secondary" variant="contained" onClick={onClick} disabled={disabled}>
-        Select
-      </Button>
+      <Tooltip title={(disabled && tooltipText) ?? ""}>
```
vinceau, line 35: Why is this necessary to check for disabled? Why can't we just pass the tooltip text straight in?

```suggestion
      <Tooltip title={tooltipText}>
```

src/settings/settingsManager.ts
```diff
@@ -45,12 +46,26 @@ export class SettingsManager {
     }
   }
 
+  public getRootSlpPath(): string {
+    return this.appSettings.settings?.rootSlpPath as string;
```
vinceau, line 50: If you're needing to cast this, you're doing something wrong. You should actually handle the `.settings` case being undefined and not use optional chaining.

src/settings/settingsManager.ts
```diff
@@ -45,12 +46,26 @@ export class SettingsManager {
...
+  }
+
+  public getUseMonthlySubfolders(): boolean {
+    return this.appSettings.settings?.useMonthlySubfolders as boolean;
```
vinceau, line 54: Again you should be handling the undefined case instead of using optional chaining and type casting.

vinceau (changes requested):

src/renderer/components/PathInput.tsx
NikhilNarayana: right that tooltip is kinda specific for the disabled case. I wanted to move it outside the component but that didn't work cause it would be on the entire component. We could just show the tooltip at all times I guess.

vinceau: Couldn't the component passing in the `tooltipText` to `PathInput` pass in undefined if not `disabled` then?

NikhilNarayana: can't be undefined, but yeah it can pass in an empty string

src/renderer/components/PathInput.tsx
```diff
@@ -30,9 +32,13 @@ export const PathInput = React.forwardRef<HTMLInputElement, PathInputProps>((pro
         <CustomInput inputRef={ref} disabled={true} value={value} placeholder={placeholder} />
         {endAdornment}
       </InputContainer>
-      <Button color="secondary" variant="contained" onClick={onClick} disabled={disabled}>
-        Select
-      </Button>
+      <Tooltip title={tooltipText ?? ""}>
+        <span>
```
vinceau, line 36: Why do we need to wrap the `Button` in a span?

vinceau (changes requested):

src/renderer/components/PathInput.tsx
NikhilNarayana: I think it could be a div or a span, but the reason is because if the Button is disabled the tooltip won't show

cnkeats: This is correct. The tooltip will also be disabled if it's immediate child is disabled. Adding an empty wrapper element is the suggested solution.

https://mui.com/components/tooltips/

vinceau (approved):
LGTM

PR #215, opened 2021-08-12
macOS child process deadlock workaround

ryanmcgrath:
On macOS, running Slippi Dolphin as a child process via Node seems to cause deadlocking issues when running the game. This was discovered during the `2.3.2/2.3.6` Slippi Dolphin release, where some improvements to the Vulkan (Metal) driver had to be rolled back due to hanging on start. After researching this further, I found I'm able to reproduce this in OpenGL as well, and with Slippi Dolphin `2.3.1` set to Vulkan... so I don't _think_ it was those changes.

If I had to theorize, it's some oddity with Node's child process module not separating the event loop for the processes on macOS. At any rate, this PR has a workaround (not a long term solution, but does make the Launcher a better experience on macOS).

- `dolphin/manager.ts` and `dolphin/instance.ts` now pass through the macOS `open` command, which ensures it's separate and can do its own thing.
- `dolphin/manager.ts` avoids tracking open Dolphin instances on macOS. This isn't really necessary there as `open` will pop a user back to the currently open `(netplay|playback)` build. It does mean that users on macOS can only watch one replay at a time, but given that limitation versus being able to play without issue... I'd take that compromise.
- macOS avoids attaching `close` event handlers since it can't own the process this way, but in practice this isn't too bad. It loses the `error` handler as well, but this might be fine to keep actually - it'd just be an error for `open` instead of the playback instance itself (so it'd catch cases where the app is inexplicably missing).
- Playback comms files are written to a subfolder in the `temp` directory, which gets cleared out and recreated on app startup. This is due to macOS needing a way to "reliably" clean these, since it can't auto clean them on Dolphin close like Windows/Linux can.

This should also head off any issues with bumping Electron and running as a universal/native app on M1 processors - without this change, I'm fairly certain we'd need to make changes to check the architecture and start Dolphin explicitly with Rosetta 2... but by passing through `open`, the system can just do whatever it needs to.

I believe this should also fix the issues some users have reported where Slippi Dolphin stuttered more when running via the Launcher, but that's not confirmed without more user testing.

I've tested this on my machines by building a local copy of Slippi Dolphin with the Metal changes that were reverted, and playback no longer hangs on a black screen. I can round up some more tests for macOS if need be.

### DIFF
--- a/src/dolphin/instance.ts
+++ b/src/dolphin/instance.ts
@@ -1,20 +1,19 @@
 import { ChildProcessWithoutNullStreams, spawn } from "child_process";
 import { randomBytes } from "crypto";
-import { app } from "electron";
 import log from "electron-log";
 import { EventEmitter } from "events";
 import * as fs from "fs-extra";
 import path from "path";
 
 import { fileExists } from "../main/fileExists";
+import { isMac } from "../common/constants";
 import { ReplayCommunication } from "./types";
+import { generateTempCommsDirectoryPath } from "../main/util";
 
 const generateTempCommunicationFile = (): string => {
-  const tmpDir = app.getPath("temp");
+  const commsDirectory = generateTempCommsDirectoryPath();
   const uniqueId = randomBytes(12).toString("hex");
-  const commFileName = `slippi-comms-${uniqueId}.json`;
-  const commFileFullPath = path.join(tmpDir, commFileName);
-  return commFileFullPath;
+  return path.join(commsDirectory, `${uniqueId}.json`);
 };
 
 export class DolphinInstance extends EventEmitter {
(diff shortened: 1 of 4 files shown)


vinceau:
How would this change affect broadcasting? Currently we automatically stop broadcasting game info once we detect the netplay Dolphin has closed.

ryanmcgrath:
Well, since you've made me think of it, I'd be surprised if broadcasting currently works at all with this PR. Game info is presumably sent over stdout or whatever to the Launcher, yeah? It's not gonna get any of that with being passed through `open`.

vinceau:
> Well, since you've made me think of it, I'd be surprised if broadcasting currently works at all with this PR. Game info is presumably sent over stdout or whatever to the Launcher, yeah? It's not gonna get any of that with being passed through `open`.

Game info is sent over enet.

PR #216, opened 2021-08-13
macOS QoL fixes

ryanmcgrath:
- Fun fact: this was somehow getting stuck with using the OpenGL software renderer. I was wondering why the hell it was so laggy on my machines, turns out if it requests Metal from Electron/Chromium it'll try to use that. Notably snappier on my 2015 MBP.

- Implements a more platform-respecting menu. Now macOS users can close the Launcher window while playing but keep the Launcher app open in the background, which is helpful for freeing up resources. Users can bring the window back with  the standard Apple-0 shortcut. Also adds a `File > Open Slippi Replay` option. Preferences are also bound to the platform hotkey for them.

- Opts macOS into the correct Big Sur styled titlebar. A few spots needed some CSS adjusting, but otherwise this fits the platform much better now. I shifted the Slippi logo from the left an additional 5px to have things line up properly.

- Changed the category identifier to be for games; this is minor but users who look in Finder > Applications categorized view will now have Slippi Launcher show up as it should.

- The `fileAssociation` on macOS was incorrectly lowercased, and notably my file associations didn't work until this was fixed. I get the sense that this isn't universal or it would've been reported (maybe?) but the correct `Info.plist` type is `Viewer` anyway, so... yeah.

Here's a screenshot of the Big Sur style window:

Screen Shot 2021-08-12 at 17 55 44

Screen Shot 2021-08-12 at 17 55 24


src/renderer/styles/platformTitleBarStyles.ts
```diff
@@ -0,0 +1,12 @@
+import { css } from "@emotion/react";
+
+// Some platforms, like macOS, need to account for extra padding at the top of the window for
+// the proper OS styling.
+export const platformTitleBarStyles =
+  process.platform === "darwin"
```
vinceau, line 6: This could just use `isMac`.

src/renderer/containers/Header/index.tsx
```diff
@@ -97,12 +98,14 @@ export const Header: React.FC<HeaderProps> = ({ path, menuItems }) => {
       css={css`
         display: flex;
         justify-content: space-between;
+        ${platformTitleBarStyles}
```
vinceau, line 101: Not really a fan of this style in the main view. It makes the header really fat and chonky. Can you maybe reduce the size of the slippi logo or something to make the header less tall?

vinceau (changes requested):

src/renderer/containers/Header/index.tsx
ryanmcgrath: macOS Big Sur styles are chonky and make use of space. Unless there's a specific reason to avoid it I'd prefer keeping it. You also need to do minimum 32px buffer on top with Electron to keep the titlebar dragging working.

src/main/index.ts
```diff
@@ -17,6 +17,12 @@ import { fileExists } from "./fileExists";
 import { setupListeners } from "./listeners";
 import { menu } from "./menu";
 
+// On macOS, we need to force Electron to use Metal if possible. Without this flag, OpenGL will be used...
+// in software rendering mode. This has a notable impact on animations on Catalina and Big Sur.
+if (isMac) {
+  app.commandLine.appendSwitch("enable-features", "Metal");
```
NikhilNarayana, line 23: Is this fine for High Sierra users?

ryanmcgrath: I believe it should be, since it's ultimately telling Chromium to use Metal if it can - and I trust Chrome to get that part right across platforms.

That said, I do not have a High Sierra test machine to try this on... and frankly that's unsupported by Apple. I understand wanting to match what mainline Dolphin supports with regards to Ishiiruka, but I'm not interested in dealing with High Sierra here - so someone else can do that legwork if it matters enough.

src/renderer/containers/Header/index.tsx
ryanmcgrath: Looking back over this, I'm still thinking it should stay as-is. Fullscreening the app and having a thinner bar up top gets odd because you can't move the mouse long in that area without the system menubar coming down and covering a chunk of it all. You'd also ideally need to adjust padding different for a fullscreen one since the traffic buttons are in a different spot there.

Electron makes this more annoying; a standard NSToolbar would have the correct behavior built-in to "just work", but I don't think this is worth dealing with.

src/renderer/styles/platformTitleBarStyles.ts
ryanmcgrath: Changed.

src/main/index.ts
NikhilNarayana: i finally found the sections of code that figure out if Metal should be used and it does handle checking if metal is supported falling back if it won't work.

NikhilNarayana:
Windows Issue: Not sure why yet, but some processes get stuck after closing the window when using this PR. This prevents the launcher from opening again until you kill those processes.

ryanmcgrath:
Per discussion w/ Nikki this now includes the Electron bump. Tested it on my M1 MBP and it runs very well. Of note:

- This checks and doesn't pass the `enable=Metal` flag to Electron if it's on an M1 device. It causes freaked out rendering for some reason, and OpenGL on M1 devices is just a shim for Metal anyway, so there's no big deal here that I can see. The flag seems to make a difference for Intel Macs rendering so would like to test it further and keep it there.

- `execFile` does seem to transparently get Rosetta 2 into the mix, which surprises me. Netplay and Playback opened fine, even when running the Launcher in `arm64`. I'll sanity check it in the morning to make sure my eyes don't deceive me, but I think this part is fine.

ryanmcgrath:
I've also added a one-line CSS addition to `styles.scss` to make font rendering look correct on macOS. Light text on darker backgrounds has bleed issues without this flag due to the way macOS handles font rendering. To my knowledge, this flag is only respected on macOS to begin with, but I welcome testing on Windows/Linux.

Before:
Screen Shot 2021-08-17 at 01 29 56

After:
Screen Shot 2021-08-17 at 01 28 02

ryanmcgrath:
> Windows Issue: Not sure why yet, but some processes get stuck after closing the window when using this PR. This prevents the launcher from opening again until you kill those processes.

For this... no idea. I remember when looking through the slow window reopen issue that this was half the issue, though, with the other half being changing the `show`/`focus` call locations. Maybe we just revert the update check but keep the `isMac`'d call locations in `main/index`?

vinceau:
I would prefer if you kept the Electron upgrade to a separate PR and not in this QoL PR. Not only because upgrading Electron needs more thought and consideration behind it, but also cuz your latest commit f663efa completely breaks the app on my MBP 16 inch running Big Sur.

Screen Shot 2021-08-18 at 5 33 15 pm

vinceau (approved):
LGTM

ryanmcgrath:
A heads up - when the Electron part of this was reverted, you also reverted the check to avoid the enableMetal flag on M1 devices. Building the main branch now bugs out for that architecture, may wanna add it back.

PR #218, opened 2021-08-13
Make slippi logo do a barrel roll on hover

rapito:
# It's quiet.. too quiet...
giphy

https://user-images.githubusercontent.com/1380104/129359530-2b834eda-3f0a-4a93-b352-5510408d7142.mov

### DIFF
--- a/src/renderer/components/BouncingSlippiLogo.tsx
+++ b/src/renderer/components/BouncingSlippiLogo.tsx
@@ -11,6 +11,11 @@ const bounceAnimation = keyframes`
   100%     { bottom: 0; }
 `;
 
+const barrelRollAnimation = keyframes`
+  0%   { transform: rotate(0); }
+  100% { transform: rotate(720deg); }
+`;
+
 const Outer = styled.div<{
   size: string;
 }>`
@@ -27,10 +32,15 @@ const Logo = styled.div<{
   background-image: url("${slippiLogo}");
   background-size: contain;
   background-repeat: no-repeat;
-  animation: ${bounceAnimation} 1.2s infinite;
+  animation: ${bounceAnimation} 1.2s infinite forwards;
   position: absolute;
-  height: ${(props) => props.size};
+  // logo is proportionally smaller to size by a factor of 0.75
+  // so we need to scale it down to avoid pivotal animations
+  height: calc(${(props) => props.size}*0.75);
   width: ${(props) => props.size};
+  &:hover {
+    animation: ${barrelRollAnimation} 1s cubic-bezier(0.68, -0.55, 0.265, 1.55) alternate forwards;
+  }
 `;
 
 export interface BouncingSlippiLogoProps {
--- a/src/renderer/components/BouncingSlippiLogo.tsx
+++ b/src/renderer/components/BouncingSlippiLogo.tsx
@@ -1,6 +1,7 @@
 import { keyframes } from "@emotion/react";
 import styled from "@emotion/styled";
 import React from "react";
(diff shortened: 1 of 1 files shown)


vinceau:
I love the idea but I don't love the implementation. Specifically this behaviour:

https://user-images.githubusercontent.com/8384734/129431044-9821caeb-16a6-41a1-8d01-662c0fb8b3ab.mov

rapito:
> I love the idea but I don't love the implementation. Specifically this behaviour:
> 
>  Screen.Recording.2021-08-14.at.11.50.11.am.mov

I refactored it to take a different approach.

vinceau:
> > I love the idea but I don't love the implementation. Specifically this behaviour:
> > Screen.Recording.2021-08-14.at.11.50.11.am.mov
> 
> I refactored it to take a different approach.

Does the new approach address the behaviour I mentioned before? If so, could you include in the PR a screengrab/video of the new behaviour when the mouse enters and leaves before the animation completes?

rapito:
> > > I love the idea but I don't love the implementation. Specifically this behaviour:
> > > Screen.Recording.2021-08-14.at.11.50.11.am.mov
> > 
> > 
> > I refactored it to take a different approach.
> 
> Does the new approach address the behaviour I mentioned before? If so, could you include in the PR a screengrab/video of the new behaviour when the mouse enters and leaves before the animation completes?

Yes:


https://user-images.githubusercontent.com/1380104/132908692-ca12479b-a782-4314-8bd1-68b5e6f1400c.mov

It only rolls once now though. And that's fine :)

src/renderer/components/BouncingSlippiLogo.tsx
```diff
@@ -23,13 +28,18 @@ const Outer = styled.div<{
 
 const Logo = styled.div<{
   size: string;
+  roll: boolean;
```
vinceau, line 31:
```suggestion
  animationState: "running" | "paused";
```
Maybe we can just pass the animation state in?

src/renderer/components/BouncingSlippiLogo.tsx
```diff
@@ -23,13 +28,18 @@ const Outer = styled.div<{
...
 }>`
   background-image: url("${slippiLogo}");
   background-size: contain;
   background-repeat: no-repeat;
-  animation: ${bounceAnimation} 1.2s infinite;
+  animation: ${bounceAnimation} 1.2s infinite forwards,
+    ${barrelRollAnimation} 1s cubic-bezier(0.68, -0.55, 0.265, 1.55) alternate forwards;
+  animation-play-state: running, ${(props) => (props.roll ? "running" : "paused")};
```
vinceau, line 38:
```suggestion
  animation-play-state: running, ${(props) => props.animationState};
```

src/renderer/components/BouncingSlippiLogo.tsx
```diff
@@ -23,13 +28,18 @@ const Outer = styled.div<{
...
   position: absolute;
-  height: ${(props) => props.size};
+  // logo is proportionally smaller to size by a factor of 0.75
+  // so we need to scale it down to avoid pivotal animations
```
vinceau, line 41:
```suggestion
```
I think I would prefer to keep comments outside of the `css` definition itself. If you want to comment maybe add it as javascript comments above the definition.

vinceau (changes requested):

rapito:
There.

vinceau (approved):
LGTM

PR #219, opened 2021-08-13
Check for updates to background, window sooner (macOS).

ryanmcgrath:
On macOS, prior to this change creating a new main window after having
closed it would take a few seconds. Forcing this check to the background
and showing the window sooner results in a quicker feeling load process.

I believe this should result in quicker appearance on other platforms as
well, but someone would need to test there.

This can be tested as of this commit by commenting out the `app.quit()`
call on `all-windows-closed`. Combined with some other PRs in the
pipeline it should work as expected.

(A good chunk of this patch was authored by Nikki; I just swapped some
window focus calls and tested on macOS)

### DIFF
--- a/src/main/index.ts
+++ b/src/main/index.ts
@@ -102,9 +102,12 @@ function createMainWindow() {
 
   window.once("ready-to-show", () => {
     didFinishLoad = true;
-    window.show();
-    window.focus();
+    //window.show();
+    //window.focus();
   });
+
+  window.show();
+  window.focus();
 
   setupListeners();
 
--- a/src/main/index.ts
+++ b/src/main/index.ts
@@ -102,12 +102,20 @@ function createMainWindow() {
 
   window.once("ready-to-show", () => {
     didFinishLoad = true;
-    //window.show();
-    //window.focus();
+
+    // Avoid paint flash in Windows by showing when ready.
+    if (!isMac) {
+      window.show();
+      window.focus();
+    }
   });
 
-  window.show();
-  window.focus();
+  // macOS doesn't seem to suffer from the paint flash, and since a user can close and reopen the
+  // window without killing the process, this should be a bit faster.
+  if (isMac) {
+    window.show();
+    window.focus();
+  }
 
   setupListeners();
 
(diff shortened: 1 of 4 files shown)


src/main/index.ts
```diff
@@ -91,10 +91,13 @@ function createMainWindow() {
...
   window.once("ready-to-show", () => {
     didFinishLoad = true;
-    window.show();
-    window.focus();
+    //window.show();
+    //window.focus();
   });
 
+  window.show();
+  window.focus();
```
NikhilNarayana, line 99: Why move these out of the ready-to-show event? Isn't that what the event is meant for?

ryanmcgrath: The window itself can show and things pop in as they're ready. I kept the `didFinishLoad` param for the place where something is waiting for the ready event; this functionally means the window shows quickly -> app loads and renders in.

It's a small distinction but feels much quicker.

ryanmcgrath: To add: I anticipate on Windows/Linux, where the _window_ represents the _app_, this isn't felt too much - sometimes apps take time to open. On macOS, reopening a window should feel quick.

ryanmcgrath: (There's no odd paint flash as `backgroundColor` is passed to the window constructor, so that helps alleviate the common case where you would need to show the window after this event)

NikhilNarayana: Paint flash happens on windows so we should keep it in ready-to-show if not on macOS.

ryanmcgrath: Kk, done.

src/main/index.ts
```diff
@@ -102,7 +113,7 @@ function createMainWindow() {
 
 // quit application when all windows are closed
 app.on("window-all-closed", () => {
-  app.quit();
+  //app.quit();
```
vinceau, line 116: Shouldn't this also have a `!isMac` check?

ryanmcgrath: This was branched from the main branch - the macOS QoL PR handles implementing window reopening, so I felt no need to deal with it here. Less moving parts.

src/main/index.ts
```diff
@@ -123,9 +134,9 @@ app.on("window-all-closed", () => {
       macMenuItem.visible = true;
     }
     return;
+  } else {
+    app.quit();
```
vinceau, line 138: Ah since the previous clause actually returns, we don't need the else statement. It's fine to just keep it as `app.quit()` without the else clause.

vinceau (changes requested):

vinceau (approved):
LGTM

PR #210, opened 2021-08-01
Add option for installing GC drivers on app install

vinceau:
This PR adds a custom installer page for optionally installing GC drivers.

[installer screenshot](https://i.imgur.com/TPwBA4E.png)

PR #221, opened 2021-08-24
feat/driver install button

L-french:
Adds a page to the launcher settings for GCC adapter driver installation. This is automated for windows, but mac and linux users will be redirected to instructions for manual installation. This is a draft PR while I write the auto-install logic.

See #172 for details.


NikhilNarayana:
I do have to note that installation is now possible at launcher install time thanks to #210, I'll have to discuss with vince if we want add an option to do so in the launcher itself at this point.

PR #224, opened 2021-08-29
feat: add toast when mirroring errors are emitted

NikhilNarayana:


src/console/workerInterface.ts
```diff
@@ -20,6 +20,8 @@ export const mirrorWorker: Promise<Thread & MirrorWorkerMethods> = new Promise((
       });
       worker.getErrorObservable().subscribe((errorMessage) => {
         mirrorLog.error(errorMessage);
+        const message = errorMessage instanceof Error ? errorMessage.message : errorMessage;
+        ipc_consoleMirrorErrorMessageEvent.main?.trigger({ message }).catch(mirrorLog.error);
```
vinceau, line 24:
```suggestion
      worker.getErrorObservable().subscribe((err) => {
        mirrorLog.error(err);
        const message = err.message ?? JSON.stringify(err);
        void ipc_consoleMirrorErrorMessageEvent.main!.trigger({ message });
```

I actually think using `void` when sending messages is pretty okay since if it errors, it's usually an error outside our control. Likewise asserting that the main attribute exists when calling from the main thread.
And then the `err.message ?? JSON.stringify(err)` pattern is used elsewhere in the codebase so we should stick with that instead of switching on `instanceOf Error`.

vinceau (changes requested):

src/renderer/containers/Console/index.tsx
```diff
@@ -74,6 +74,15 @@ export const Console: React.FC = () => {
     // Close the dialog
     onCancel();
   };
+
+  ipc_consoleMirrorErrorMessageEvent.renderer!.handle(async ({ message }) => {
```
vinceau, line 78: This should be `useEvent` instead of `handle` since `handle` expects the renderer to clean up and unsubscribe.

Also you should consider when we want to unsubscribe. If you add this line of code inside of the `Console` component, if you happen to have browsed to a different page, it would have unsubscribed already so you miss the messages. Global events such as these which should be subscribed to regardless of current page should probably be moved to `useAppListeners`.

vinceau (changes requested):

src/console/workerInterface.ts
NikhilNarayana: so the reason I use the instanceof method is because we have a string | Error type, so we have to do a type check.

vinceau: In that case that's fine to use the `instanceof`. But I would still prefer the optional chaining `?` to be replaced with non-null assertion `!`.

vinceau (changes requested):

src/console/workerInterface.ts
NikhilNarayana: whoops, i didn't even notice that I used an optional chaining there, fixed.

vinceau (approved):
LGTM

PR #225, opened 2021-08-29
fix: refresh broadcast/spectate auth token

NikhilNarayana:


src/broadcast/types.ts
```diff
@@ -22,13 +22,15 @@ export enum BroadcastEvent {
   DOLPHIN_STATUS_CHANGE = "DOLPHIN_STATUS_CHANGE",
   ERROR = "ERROR",
   LOG = "LOG",
+  RECONNECT_1006 = "RECONNECT_1006",
```
vinceau, line 25: I'm not sure what the 1006 number is referring to, probably the specific error code that it returns, but as a name, it's not very useful. Like seeing `RECONNECT_1006` isn't very descriptive. Can we rename this to something that is more clearer to what it represents. Such that when someone sees that there is a susbcription to the `RECONNECT_1006` event, they would know have a better understanding about when/what it means when it gets triggered.

This goes for all references to 1006 in this PR. Like unless 1006 is some universal number meaning some specific instance of reconnection that I'm out of the loop on, I think it would be best if references to this 1006 number was something more descriptive.

vinceau (changes requested):

src/renderer/lib/hooks/useAppListeners.ts
```diff
@@ -193,4 +198,14 @@ export const useAppListeners = () => {
...
   }, [rootSlpPath, extraSlpPaths, init]);
+
+  const [startBroadcast] = useBroadcast();
+  ipc_broadcastReconnect.renderer!.useEvent(async ({ config }) => {
+    startBroadcast(config).catch(log.error);
+  }, []);
+
+  const [, refreshBroadcasts] = useBroadcastList();
+  ipc_spectateReconnect.renderer!.useEvent(async () => {
+    refreshBroadcasts()?.catch(log.error);
```
vinceau, line 209: Why do you have an optional chain here?

```suggestion
    refreshBroadcasts().catch(log.error);
```

src/broadcast/spectateWorker.ts
```diff
@@ -47,6 +49,10 @@ spectateManager.on(SpectateEvent.NEW_FILE, async (playbackId: string, filePath:
   spectateDetailsSubject.next({ playbackId, filePath });
 });
 
+spectateManager.on(SpectateEvent.RECONNECT, async () => {
+  reconnectSubject.next(1);
```
vinceau, line 53: Is there a reason why we return a number here? Does it mean anything?

Because otherwise if the return value means nothing, we can just make it an `Observable<void>` and simply call `.next()`.

vinceau (changes requested):

src/broadcast/spectateWorker.ts
NikhilNarayana: didn't realize we could do that, will fix.

src/renderer/lib/hooks/useAppListeners.ts
NikhilNarayana: it is a `Promise<void> | undefined` type, I could do a quick check.

vinceau (dismissed):
LGTM, but it would be good if you added some more documentation somewhere like in this PR description, and/or in the codebase mentioning the token expiring after an hour and why we have to manually handle this case.

vinceau (approved):
LGTM

PR #227, opened 2021-09-01
feat: use log scopes everywhere

NikhilNarayana:


src/main/index.ts
```diff
@@ -28,6 +28,7 @@ if (isMac && process.arch !== "arm64") {
 }
 
 // use console.log as log.debug for easier access to debug logging
+const log = electronLog.scope("devDebug");
```
vinceau, line 31: Why is this called `devDebug`?

src/renderer/index.tsx
```diff
@@ -7,6 +7,7 @@ import App from "./App";
 const rootEl = document.getElementById("app");
 
 // use console.log as log.debug for easier access to debug logging
+const log = electronLog.scope("devDebug");
```
vinceau, line 10: Here's another `devDebug`?

vinceau (changes requested):

vinceau (approved):
LGTM

PR #229, opened 2021-09-04
fix: sticky shift key bug when starting a recording with ShareX

JLaferri:
This could have further implications for `useMouseTrap`. It seems like under certain conditions, the program may not pick up a key making it a pretty bad choice for managing an "isShiftHeld" state.

In this case, I was activating my ShareX recording software which records the screen by pressing `Shift + PrntScrn`. I was not pressing shift any other time. My guess is because ShareX freezes the screen when taking a screenshot, electron does not pick up the release of the key? May be other scenarios where this can happen though.

https://user-images.githubusercontent.com/1534726/132077505-38455ce4-5170-45ac-9c9c-e320f25dc245.mp4


vinceau (approved):
LGTM. Didn't know about `event.shiftKey`. That's neat.

PR #230, opened 2021-09-04
feat: add useDolphin hook

NikhilNarayana:
This PR is meant to add a hook that UI components can use to open dolphin and keep track of if dolphin is open.


src/renderer/lib/hooks/useDolphin.ts
```diff
@@ -0,0 +1,103 @@
...
+      },
+    }),
+  ),
+);
+
+export const useDolphin = () => {
+  const { addToast } = useToasts();
+  const setDolphinOpen = useDolphinStore((store) => store.setDolphinOpen);
+
+  const handleResult = (res: MainEndpointResponse<any>, errLog: string): boolean => {
```
vinceau, line 41:
```suggestion
  const handleResult = <T>(res: MainEndpointResponse<T>, errLog: string): boolean => {
```

Should probably just use a generic for this.

src/renderer/lib/hooks/useDolphin.ts
```diff
@@ -0,0 +1,103 @@
+/* eslint-disable react-hooks/rules-of-hooks */
```
vinceau, line 1: Why are these eslint rules disabled?

vinceau (changes requested):
The PR itself looks okay but I'm still not sure I understand the PR in the context of the PR description. Like do your changes actually fix the issue you're describing? Or does it not actually fix the problem? That's not clear in the description or the code.

src/renderer/lib/hooks/useDolphin.ts
NikhilNarayana: that's a leftover from when I was trying to write the hook in a different way, I'll remove it.

NikhilNarayana:
> The PR itself looks okay but I'm still not sure I understand the PR in the context of the PR description. Like do your changes actually fix the issue you're describing? Or does it not actually fix the problem? That's not clear in the description or the code.

the problem is not fixed. I need to investigate a # bit further to find the true root cause.

vinceau (approved):
Conditional approval, can you:
1. Move the issue you're describing into a Github issue
2. Update this PR description with context/justification for the actual PR changes

PR #232, opened 2021-09-05
Move component IPC calls to library

### LINKED ISSUE #183: move IPC calls to functions in lib
Components should not have to deal with handling IPC calls directly. IPC calls should be encapsulated inside of an async function inside of `lib` which components can then import.

It may also be helpful to have the functions exported inside of a hook so that the exported function already handles errors using `addToast`.

L-french:
Refactors two components which directly handled IPC calls, moving the IPC stuff to `lib` files. Covers some of #183, including the error handling mentioned in the issue.


src/renderer/lib/hooks/useHandleDragStart.ts
```diff
@@ -0,0 +1,23 @@
+import { ipcRenderer } from "electron";
+import { useToasts } from "react-toast-notifications";
+
+export const useHandleDragStart = () => {
```
vinceau, line 4: The hook should represent what it's actually used for. In this case a more suitable name would be something like `useFileDrag`.

```suggestion
export const useFileDrag = () => {
```

src/renderer/lib/hooks/useHandleDragStart.ts
```diff
@@ -0,0 +1,23 @@
...
+  const { addToast } = useToasts();
+
+  const handleDragStart = async (e: React.DragEvent<HTMLDivElement>, filePaths: string[]) => {
```
vinceau, line 7: It's best not to use single letter variable names for parameters.

```suggestion
  const handleDragStart = async (event: React.DragEvent<HTMLDivElement>, filePaths: string[]) => {
```

src/renderer/lib/hooks/useUpdateHandler.ts
```diff
@@ -0,0 +1,20 @@
+import { ipc_installUpdate } from "common/ipc";
+import { useToasts } from "react-toast-notifications";
+
+export const useUpdateHandler = () => {
```
vinceau, line 4: Likewise this should be renamed to something like: `useInstallAppUpdate` or something.

```suggestion
export const useInstallAppUpdate = () => {
```

vinceau (changes requested):
Not bad! Just needs some minor changes.

L-french:
> Not bad! Just needs some minor changes.

No problem--I preserved old variable names where I could, but it makes sense to adjust them for clarity while we're here. Changes made in c9edb0faa080689c486e0f60e71bb53b3aef98af.

vinceau (approved):
LGTM! Nice work!

Just addressing the previous comment, it's not that bad if a local function definition has poorly named function parameters, but if the function is exported, it's important for those parameters to have clear names so it's obvious what needs to be passed in.

PR #236, opened 2021-09-16
feat: mirroring QoL improvements

NikhilNarayana:
* make mirror settings viewable even when connected so streamers can verify that the settings are correct without disconnecting from the wii.
* add the console nick to the folder path
  * image

### DIFF
--- a/src/console/mirrorManager.ts
+++ b/src/console/mirrorManager.ts
@@ -90,11 +90,12 @@ export class MirrorManager extends EventEmitter {
       connection.on(ConnectionEvent.HANDSHAKE, (details: ConnectionDetails) => {
         this.emit(MirrorEvent.LOG, "Got handshake from wii");
         this.emit(MirrorEvent.LOG, details);
+        if (config.useNicknameFolders) {
+          const replayFolder = path.join(config.folderPath, details.consoleNick);
+          fs.ensureDirSync(replayFolder);
 
-        const replayFolder = path.join(config.folderPath, details.consoleNick);
-        fs.ensureDirSync(replayFolder);
-
-        fileWriter.updateSettings({ consoleNickname: details.consoleNick, folderPath: replayFolder });
+          fileWriter.updateSettings({ consoleNickname: details.consoleNick, folderPath: replayFolder });
+        }
 
         this.emit(MirrorEvent.MIRROR_STATUS_CHANGE, {
           ip: config.ipAddress,
--- a/src/console/mirrorManager.ts
+++ b/src/console/mirrorManager.ts
@@ -90,7 +90,11 @@ export class MirrorManager extends EventEmitter {
       connection.on(ConnectionEvent.HANDSHAKE, (details: ConnectionDetails) => {
         this.emit(MirrorEvent.LOG, "Got handshake from wii");
         this.emit(MirrorEvent.LOG, details);
-        fileWriter.updateSettings({ consoleNickname: details.consoleNick });
+
+        const replayFolder = path.join(config.folderPath, details.consoleNick);
+        fs.ensureDirSync(replayFolder);
+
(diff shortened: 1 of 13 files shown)


src/renderer/containers/Console/AddConnectionDialog.tsx
```diff
@@ -19,13 +19,15 @@ export interface AddConnectionDialogProps {
   selectedConnection: Partial<StoredConnection> | null;
   onSubmit: (conn: Omit<StoredConnection, "id">) => void;
   onCancel: () => void;
+  disableForm: boolean;
```
vinceau, line 22: Can we get this to just be disabled too?

```suggestion
  disabled?: boolean;
```

src/renderer/containers/Console/AddConnectionForm.tsx
```diff
@@ -32,9 +32,10 @@ type FormValues = {
 export interface AddConnectionFormProps {
   defaultValues?: Partial<FormValues>;
   onSubmit: (values: FormValues) => void;
+  disableForm: boolean;
```
vinceau, line 35:
```suggestion
  disabled?: boolean;
```

src/renderer/containers/Console/SavedConnectionItem.tsx
```diff
@@ -100,7 +96,7 @@ export const SavedConnectionItem: React.FC<SavedConnectionItemProps> = ({
                 font-size: 14px;
               `}
             >
-              {connection.folderPath}
+              {path.join(connection.folderPath, nickname ?? "")}
```
vinceau, line 99: I'm not sure this is a good idea. Wouldn't this mean that you can't set a custom location for the files without it always having the nickname appended? e.g. if i set the `connection.folderPath` to be `/abc/test` it would always be saved to `/abc/test/nickname`? I'm not sure I'm not board with forcing that,

src/console/mirrorManager.ts
```diff
@@ -90,7 +90,11 @@ export class MirrorManager extends EventEmitter {
       connection.on(ConnectionEvent.HANDSHAKE, (details: ConnectionDetails) => {
         this.emit(MirrorEvent.LOG, "Got handshake from wii");
         this.emit(MirrorEvent.LOG, details);
-        fileWriter.updateSettings({ consoleNickname: details.consoleNick });
+
+        const replayFolder = path.join(config.folderPath, details.consoleNick);
+        fs.ensureDirSync(replayFolder);
```
vinceau, line 95: I'm not sure I like the idea of forcing the file to always be in the nickname folder. I think it should be an option but not mandated. Can we come up with an alternative option?

vinceau (changes requested):

src/renderer/containers/Console/SavedConnectionItem.tsx
NikhilNarayana: I need some better justification about why that is users might not want them sorted by nickname because I can't see any true downsides. If we have a strong justification then my next question is do we want this to be a setting that gets applied to all consoles or individually?

vinceau: I guess for starters, if I choose a folder to be the folder path for the output files, I would expect the files to be output there -- not in a subdirectory. At the moment in this implementation, I don't think that it's clear that it will end up being saved in the subdirectory. You could potentially address this by heavy signposting. But even then, with the current implementation, the file path should be set at a higher level.  i.e. if (as in the example above) you need to do a `path.join()` at the renderer level, something is wrong. The `connection.folderPath` should have already had the nickname appended by the time it reaches that stage.

Personally speaking I would rather give the users options rather than forcing a particular behaviour. Whether that's a global option in settings that says "Save all spectator games into individual folder", or a checkbox when adding a console, or even path templating like the OBS output file path, I really don't mind.

vinceau (approved):

PR #237, opened 2021-09-19
fix: mirroring failures caused by use of OBS websocket

NikhilNarayana:
This PR attempts to fix 2 known bugs surrounding the auto switcher.

Current Behavior
1. If OBS is not open, attempting to mirror will provide a toast that just says `some stuff [object Object]`, which is useless to the end user
2. If OBS closes while the auto switcher is connected, when the auto switcher triggers the UI will crash.

### DIFF
--- a/src/console/autoSwitcher.ts
+++ b/src/console/autoSwitcher.ts
@@ -57,13 +57,11 @@ export class AutoSwitcher extends EventEmitter {
           password: this.obsPassword,
         });
       } catch (err) {
-        if (err) {
-          this.emit(
-            MirrorEvent.ERROR,
-            "Could not connect to OBS, ensure you have OBS websocket installed and OBS is open.",
-          );
-          return;
-        }
+        this.emit(
+          MirrorEvent.ERROR,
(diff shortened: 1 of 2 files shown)


src/console/autoSwitcher.ts
```diff
@@ -51,17 +51,21 @@ export class AutoSwitcher extends EventEmitter {
...
-        },
-        (err) => {
-          if (err) {
-            this.emit(MirrorEvent.ERROR, err);
-          }
-        },
-      );
+        });
+      } catch (err) {
+        if (err) {
```
vinceau, line 60: Is it possible for an error to be thrown with `err` being falsey? Cuz if it isn't then we shouldn't need to check for `err` existing. If it is possible for it to be falsy, can we reverse the check to save an indentation level?

```suggestion
        if (!err) {
          return;
        }
```

vinceau (changes requested):

src/console/autoSwitcher.ts
NikhilNarayana: yeah this shouldn't even need a check

vinceau (approved):
LGTM

Issue #178, opened 2021-06-23
Feat: Replay DB


NikhilNarayana:
Add a database that caches replay data so we don't have to parse replay files all the time.

PR #241, opened 2021-10-10
Fast replay loading 2

Kered13:
Greatly improves the responsiveness of the replay list. To be clear, replays do not actually load faster. Instead they only appear to load faster because replays are displayed as they are ready instead of waiting for the entire replay directory to load. On my machine with ~12k replays on an SSD, it takes one minute to load the entire directory from a cold drive. With this change the first page of replays appears almost immediately (less than one second).

This is achieved by:

1. Separating the loading of replay metadata (name, path, and file creation time) from loading the replay data. After the metadata is loaded replays begin loading in a separate thread and details are filled in when they become available.
2. Updates to the replay list are batched to minimize renders and improve performance. Batching parameters were tuned by experiment on my machine (~12k replays on an SSD).
3. When setting new state for replay files, the replay state map is not copied. This saves significant time each time state is updated. A dummy state is updated at the same time to force a re-render of the ReplayBrowser component.
4. Replays are loaded in the expected display order. This is achieved by using the file creation time instead of the game start time, so it may not always be accurate. Directory loading could likely be further improved if this behavior were dropped so that every file in the directory did not need to be stat'd, but then it would not be possible to prioritize loading the replays at the top of the list.

Further improvements might be possible if replays (SlippiGame objects) could be loaded asynchronously, or if a worker pool were used. UPDATE_BATCH_SIZE would then need to be increased to take advantage.

Replays that are still loading are hidden in the replay list. This can be changed in the future if desired.


src/renderer/lib/hooks/useReplays.ts
```diff
@@ -26,6 +31,11 @@ type StoreState = {
     total: number | null;
     fileResult: FileResult | null;
   };
+  // Can be incremented to force ReplayBrowser to re-render. This is useful
+  // because we can significantly improve performance by updating files in place
+  // (without copying the map), but this will not normally trigger a render
+  // because React does not recognize it as a state change.
+  forceRender: number;
```
NikhilNarayana, line 38: this feels like an anti-pattern in react, avoiding this kind of thing would be ideal.

NikhilNarayana (changes requested):
I have two main concerns: 1) the changes in useReplay that go against the react design pattern and 2) I would like to see an optimistic loading pattern, right we sit on a blank screen for a few seconds before replays even show up which feels like meh UX. 

Also please use prettier/eslint to confirm everything is actually formatted correctly.

src/renderer/lib/hooks/useReplays.ts
Kered13: The problem is that doing a normal state update causes the entire files map to be copied, which becomes incredibly slow when this is being done for every update. I tried to implement this normally again tonight, trying two different approaches that I hoped would be a more clever, but it was still far too slow. First I tried using Immer, this was the slowest by far, I believe because it makes a deep copy for the draft state. Then I tried using Zustand alone, writing the code to intentionally avoid the deep copy, but I still had to make a shallow copy of the map and it was still slow.

The root of the issue as far as I know is React's insistence on immutable data. If you have an immutable map of 12k files (which is the number of replay files in my folder right now) and update one of those entries, best case scenario you have to copy 12k references in addition to the actual data you are updating. Do that 12k times, once for each entry as the files finish loading, and you're copying 144,000,000 references unnecessarily. This is not trivial.

The code above is the only solution that I have found that maintains good performance: Trick React by update the map in place, avoiding any unnecessary copies, and then updating the signal variable to tell React to start a new render.

If you have other ideas I'm willing to try them out.

src/replays/main.ts
```diff
@@ -1,18 +1,41 @@
...
+  return await w.loadReplayFolder(folderPath);
+});
+
+let hasStartedObservers = false;
+ipc_loadReplayFiles.main!.handle(async ({ fileHeaders, batcherId }) => {
+  const w = await replayBrowserWorker;
+  // We only need to do this once. These observers never get unsubscribed, so if
+  // we did it more we would leak memory.
+  if (!hasStartedObservers) {
+    w.getFileLoadCompleteObservable().subscribe((fileLoadComplete) => {
```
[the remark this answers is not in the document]
Kered13: Good point, done!

Kered13:
All comments have been addressed, and I merged the latest main branch to head off merge conflicts (there was only one minor conflict which I have resolved).

vinceau:
I'm not really keen to merge this PR for a few reasons:
1. As you said yourself: "replays do not actually load faster. Instead they only appear to load faster". The issue here is that you're showing an intermediate state as though it is the final state. You can experience this yourself by putting a term in the search bar, and then changing replay directories. Instead of just showing, loading bar -> no results found. it would show a whole bunch of spinning replays -> no results found.
2. "Replays are loaded in the expected display order". It was a conscious design decision to use the timestamp as reported by the SLP file. If we lift this requirement then yes replay loading would be much much faster. We might want to reconsider this design decision at a later time since the Wii clocks especially can be unreliable, but for the time being we've decided to stick with this design.
3. This if anything is a bandaid fix with the real solution being to implement #178. In preparation for the replay db the replay loading code is going to drastically change anyway. So merging this PR despite everything mentioned above would be additional code that is likely going to be changed/modified anyway.

Kered13:
1. This is easily fixed by not showing loading spinners for replays that are still being loaded. However showing the loading spinners was explicitly requested when I brought this issue up on the Discord. Personally I can go either way, but this is not a blocking issue, only a question of what the team prefers.
2. The vast majority of users are dealing with slippi replays that were recorded and saves on PCs with accurate clocks. This should not be considered a blocking issue, dealing with replays from Wiis is a minor use case.
3. There has been absolutely no progress on this bug in over a year, nor any signs of progress. No one is actively working on it, unless they are doing it in secret, nor are there any concrete designs for it as far as I can tell. Therefore we cannot expect that it will be done any time soon. On the other hand, I have a solution that dramatically improves loading performance _right now_ that works _very_ well. Furthermore, my solution is not in any way incompatible with the proposed database approach, nor do my code changes do anything to get in the way of a future database implementation. In fact, my solution will _still_ complement the database solution, because the database solution does not work when large numbers of new replays are added at once, the database can only work for replays that have already been cached. However my solution will improve performance for all replays.

To be completely frank, the idea of loading and parsing _every single replay in a folder_ at once before showing any results was ridiculously bad to begin with. I'm honestly shocked it was ever seriously considered, it's obviously unscalable. But I have fixed the problem for you.

PR #256, opened 2021-12-11
feat: client side display name validation

NikhilNarayana:
The backend and website include the same validation.

### DIFF
--- a/src/renderer/containers/Header/NameChangeDialog.tsx
+++ b/src/renderer/containers/Header/NameChangeDialog.tsx
@@ -10,7 +10,7 @@ import { ConfirmationModal } from "@/components/ConfirmationModal";
 import { useAccount } from "@/lib/hooks/useAccount";
 import { useAsync } from "@/lib/hooks/useAsync";
 import { changeDisplayName } from "@/lib/slippiBackend";
-import { isValidDisplayName } from "@/lib/validate";
+import { validateDisplayName } from "@/lib/validate";
 
 export const NameChangeDialog: React.FC<{
   displayName: string;
@@ -88,7 +88,7 @@ export const NameChangeDialog: React.FC<{
               }}
             />
           )}
-          rules={{ validate: isValidDisplayName }}
+          rules={{ validate: validateDisplayName }}
         />
       </ConfirmationModal>
     </div>
--- a/src/renderer/containers/Header/NameChangeDialog.tsx
+++ b/src/renderer/containers/Header/NameChangeDialog.tsx
@@ -3,19 +3,24 @@ import { css, jsx } from "@emotion/react";
 import CircularProgress from "@material-ui/core/CircularProgress";
 import TextField from "@material-ui/core/TextField";
 import React from "react";
+import { Controller, useForm } from "react-hook-form";
 import { useToasts } from "react-toast-notifications";
 
 import { ConfirmationModal } from "@/components/ConfirmationModal";
 import { useAccount } from "@/lib/hooks/useAccount";
 import { useAsync } from "@/lib/hooks/useAsync";
 import { changeDisplayName } from "@/lib/slippiBackend";
+import { isValidDisplayName } from "@/lib/validate";
 
 export const NameChangeDialog: React.FC<{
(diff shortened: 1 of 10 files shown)


src/renderer/lib/validate.ts
```diff
@@ -45,3 +45,21 @@ export const isValidConnectCodeStart = (codeStart: string) => {
 
   return true;
 };
+
+export const isValidDisplayName = (displayName: string): string | true => {
+  displayName = displayName.trim();
```
vinceau, line 50: I'm not a big fan of modifying function parameters like this tbh. I think the caller of the function should just pass in the trimmed version of the string.

```suggestion
```

src/renderer/lib/validate.ts
```diff
@@ -45,3 +45,21 @@ export const isValidConnectCodeStart = (codeStart: string) => {
 
   return true;
 };
+
+export const isValidDisplayName = (displayName: string): string | true => {
```
vinceau, line 49: This is something that I've noticed but I would prefer for us to distinguish between an `isValidX` function (which has the type signature `() => boolean` versus a error validation (which has the type signature `() => string | true`. I think if we return the error message in the function we should call the function name `validateX` and if we simply return a boolean we should call the function name `isValidX`. I think this would make things a bit clearer.

```suggestion
export const validateDisplayName = (displayName: string): string | true => {
```

src/renderer/containers/Header/NameChangeDialog.tsx
```diff
@@ -64,15 +71,24 @@ export const NameChangeDialog: React.FC<{
...
+              required={true}
+              error={Boolean(error)}
+              helperText={error ? error.message : undefined}
+              autoFocus={true}
+              inputProps={{
+                maxLength: 15,
+              }}
+            />
+          )}
+          rules={{ validate: isValidDisplayName }}
```
vinceau, line 91:
```suggestion
          rules={{ validate: (val) => isValidDisplayName(val.trim()) }}
```

I think we should pass the trimmed version here instead of delegating that to the validation logic.

vinceau (changes requested):

src/renderer/lib/validate.ts
```diff
@@ -1,4 +1,4 @@
-export const isValidIpAddress = (ip: string) => {
+export const validateIpAddress = (ip: string): boolean => {
```
vinceau, line 1:
```suggestion
export const isValidIpAddress = (ip: string): boolean => {
```

This returns only a boolean so it should still be called `isValidIpAddress`.

src/renderer/lib/validate.ts
```diff
@@ -11,7 +11,7 @@ export const isValidIpAddress = (ip: string) => {
   return chunks.map((n) => parseInt(n)).every((n) => n >= 0 && n <= 255);
 };
 
-export const isValidIpAndPort = (ip: string) => {
+export const validateIpAndPort = (ip: string): string | boolean => {
```
vinceau, line 14:
```suggestion
export const validateIpAndPort = (ip: string): string | true => {
```

src/renderer/lib/validate.ts
```diff
@@ -23,10 +23,10 @@ export const isValidIpAndPort = (ip: string) => {
     return "Invalid Port";
   }
 
-  return isValidIpAddress(ip);
+  return validateIpAddress(ip);
```
vinceau, line 26:
```suggestion
  if (!isValidIpAddress(ip)) {
    return "Invalid IP address";
  }
  
  return true;
```

For completeness I think we should do this.

vinceau (changes requested):
Thanks for adding in those extra test cases! Just a few more minor nits and it should be good. :)

src/renderer/lib/validate.ts
NikhilNarayana: this doesn't work because isValidIpAddress returns a bool.

vinceau (approved):
LGTM, though the builld seems to be broken. :|

vinceau (approved):

PR #258, opened 2021-12-30
fix: show error when dolphin is missing dlls

JLaferri:
Handles non-zero exit codes when Dolphin closes unexpectedly by showing an error message. Prior to this nothing would happen and the user would have no idea what is happening.

image

TODO:
1. Test this on other operating systems, make sure closing the application normally doesn't cause errors messages.
2. Improve copy?

Improvements:
1. Make it so that if the error code detected is related to missing DLLs, direct the user to a download link for VCR
2. Perhaps increase the prominence of the error

### DIFF
--- a/FAQ.md
+++ b/FAQ.md
@@ -6,7 +6,7 @@ Rollback is a superior netcode implementation to the delay based netcode that ev
 
 We only support NTSC-U/J 1.02.
 
-## Nothing happens when I hit Play/Necessary DLLs are missing
+## It says "Required DLLs are missing" when I hit Play!
 
 You are missing DLLs necessary for Dolphin to run, download and install the file at the link: <https://aka.ms/vs/17/release/vc_redist.x64.exe>
 
--- a/FAQ.md
+++ b/FAQ.md
@@ -6,6 +6,10 @@ Rollback is a superior netcode implementation to the delay based netcode that ev
 
 We only support NTSC-U/J 1.02.
 
+## Nothing happens when I hit Play/Necessary DLLs are missing
+
+You are missing DLLs necessary for Dolphin to run, download and install the file at the link: <https://aka.ms/vs/17/release/vc_redist.x64.exe>
+
 ## How do I setup my GameCube Controller Adapter?
 
 Set your adapter to Wii U/Switch mode and then follow the section for your OS.
@@ -74,10 +78,6 @@ We cannot help you find an ISO, you will need to acquire one yourself. This webs
 ## Can I play Free For All with Slippi Online?
 
 No, only singles (1v1) and teams (2v2, 3v1, 2v1v1) is supported at this time. Free For All won't be added in the near future, but is not off the table entirely.
-
-## I get a VCRUNTIME error whenever I launch Dolphin
-
-Install this: <https://aka.ms/vs/17/release/vc_redist.x64.exe>
 
 ## Some of my inputs aren't going through correctly
 
--- a/src/dolphin/main.ts
+++ b/src/dolphin/main.ts
@@ -115,10 +115,10 @@ ipc_checkDesktopAppDolphin.main!.handle(async () => {
   return { dolphinPath: dolphinExecutablePath, exists: exists };
 });
 
-dolphinManager.on("playback-dolphin-closed", async () => {
-  void ipc_dolphinClosedEvent.main!.trigger({ dolphinType: DolphinLaunchType.PLAYBACK });
+dolphinManager.on("playback-dolphin-closed", async (_playbackId: string, code: number) => {
+  void ipc_dolphinClosedEvent.main!.trigger({ dolphinType: DolphinLaunchType.PLAYBACK, exitCode: code });
 });
 
-dolphinManager.on("netplay-dolphin-closed", async () => {
-  void ipc_dolphinClosedEvent.main!.trigger({ dolphinType: DolphinLaunchType.NETPLAY });
+dolphinManager.on("netplay-dolphin-closed", async (code: number) => {
+  void ipc_dolphinClosedEvent.main!.trigger({ dolphinType: DolphinLaunchType.NETPLAY, exitCode: code });
 });
--- a/src/renderer/lib/hooks/useAppListeners.ts
+++ b/src/renderer/lib/hooks/useAppListeners.ts
@@ -229,7 +229,28 @@ export const useAppListeners = () => {
 
   const setDolphinOpen = useDolphinStore((store) => store.setDolphinOpen);
   ipc_dolphinClosedEvent.renderer!.useEvent(
-    async ({ dolphinType }) => {
+    async ({ dolphinType, exitCode }) => {
+      if (exitCode === 0xc0000135) {
+        addToast(
+          `Necessary DLLs for launching Dolphin are missing. Head to the FAQ in the settings to find out how to install them.`,
+          {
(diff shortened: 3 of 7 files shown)


src/renderer/lib/hooks/useAppListeners.ts
```diff
@@ -229,7 +229,19 @@ export const useAppListeners = () => {
 
   const setDolphinOpen = useDolphinStore((store) => store.setDolphinOpen);
   ipc_dolphinClosedEvent.renderer!.useEvent(
-    async ({ dolphinType }) => {
+    async ({ dolphinType, exitCode }) => {
+      if (exitCode !== 0) {
```
NikhilNarayana, line 233: we should handle the 2 known exit codes (0xC0000135/3221225781 for vcr and 3 for dolphin updates) with this case being the final `else if`.

src/dolphin/main.ts
```diff
@@ -115,10 +115,10 @@ ipc_checkDesktopAppDolphin.main!.handle(async () => {
   return { dolphinPath: dolphinExecutablePath, exists: exists };
 });
 
-dolphinManager.on("playback-dolphin-closed", async () => {
-  void ipc_dolphinClosedEvent.main!.trigger({ dolphinType: DolphinLaunchType.PLAYBACK });
+dolphinManager.on("playback-dolphin-closed", async (playbackId: string, code: number) => {
```
NikhilNarayana, line 118:
```suggestion
dolphinManager.on("playback-dolphin-closed", async (_playbackId: string, code: number) => {
```
since it is unused

NikhilNarayana (changes requested):

src/renderer/lib/hooks/useAppListeners.ts
JLaferri: Yeah this is what I mentioned in the improvements section. I'll see about putting links in the toaster.

src/dolphin/main.ts
JLaferri: Cool thanks, I was gonna ask how to do this.

src/renderer/lib/hooks/useAppListeners.ts
```diff
@@ -229,7 +229,19 @@ export const useAppListeners = () => {
...
+        addToast(
+          `Dolphin encountered an error with code: 0x${exitCode.toString(16)}.
+          If you are having issues, please screenshot this and post it in a support channel in the
+          Slippi Discord.`,
```
vinceau, line 237:
```suggestion
          `Dolphin exited with error code: 0x${exitCode.toString(16)}. Visit the Slippi Discord for help with this error.`,
```
Thoughts on simplifying the error message copy to just this?

JLaferri: It's shorter but less clear. I think people will show up after dismissing the message and not writing down the code.

src/main/index.ts
```diff
@@ -27,6 +27,8 @@ if (isMac && process.arch !== "arm64") {
   app.commandLine.appendSwitch("enable-features", "Metal");
 }
 
+process.env.SLIPPI_LAUNCHER = "true";
+
```
vinceau, line 31: What is this and why do we need it?

src/renderer/lib/hooks/useAppListeners.ts
```diff
@@ -229,7 +229,28 @@ export const useAppListeners = () => {
 
   const setDolphinOpen = useDolphinStore((store) => store.setDolphinOpen);
   ipc_dolphinClosedEvent.renderer!.useEvent(
-    async ({ dolphinType }) => {
+    async ({ dolphinType, exitCode }) => {
+      if (exitCode === 0xc0000135) {
+        addToast(
+          `Necessary DLLs for launching Dolphin are missing. Head to the FAQ in the settings to find out how to install them.`,
```
vinceau, line 235:
```suggestion
          `Required DLLs for launching Dolphin are missing. Check the Help section in the settings page to fix this issue.`,
```

src/renderer/lib/hooks/useAppListeners.ts
```diff
@@ -229,7 +229,28 @@ export const useAppListeners = () => {
...
+            id: `dolphin-error-0x${exitCode.toString(16)}`,
+            appearance: "error",
+            autoDismiss: false,
+          },
+        );
+      } else if (exitCode) {
+        addToast(
+          `Dolphin encountered an error with code: 0x${exitCode.toString(16)}.
+          If you are having issues, please screenshot this and post it in a support channel in the
+          Slippi Discord.`,
```
vinceau, line 246:
```suggestion
          `Dolphin exited with error code: 0x${exitCode.toString(16)}.
          Please screenshot this and post it in a support channel in the Slippi Discord for assistance.`,
```

FAQ.md
```diff
@@ -6,6 +6,10 @@ Rollback is a superior netcode implementation to the delay based netcode that ev
 
 We only support NTSC-U/J 1.02.
 
+## Nothing happens when I hit Play/Necessary DLLs are missing
```
vinceau, line 9:
```suggestion
## It says "Required DLLs are missing" when I hit Play!
```

vinceau (changes requested):
Mostly minor copy changes.

src/main/index.ts
NikhilNarayana: future thing for letting dolphin know it was launched by the Launcher and that it should let the launcher handle the update process.

src/renderer/lib/hooks/useAppListeners.ts
JLaferri: I included the first part in the case where people see the error when Dolphin exits poorly when being closed (maybe threading stuff). But I don't mind dropping it if you don't mind more people going to the support channels.

src/main/index.ts
NikhilNarayana: pulled from this PR, will address in the future.

src/renderer/lib/hooks/useAppListeners.ts
NikhilNarayana: I think I would prefer any weird exit codes to be reported. I may add an exception for 3 on windows since that is known to happen when they select `Update`.

vinceau: Yeah I think if it's something they shouldn't need to worry about, we shouldn't show the toast in the first place.

src/renderer/lib/hooks/useAppListeners.ts
```diff
@@ -229,7 +230,32 @@ export const useAppListeners = () => {
 
   const setDolphinOpen = useDolphinStore((store) => store.setDolphinOpen);
   ipc_dolphinClosedEvent.renderer!.useEvent(
-    async ({ dolphinType }) => {
+    async ({ dolphinType, exitCode }) => {
+      // ignore some situations
+      const ignore = [
+        exitCode === 3 && !isLinux && !isMac, // when selecting Update in game on windows, dolphin returns 3
+      ];
```
vinceau, line 237: This logic is getting complicated now with the ignoring.

Can you create a helper function with the signature `handleDolphinExitCode(exitCode: number): string | null` so that we can simply check if it's a string and emit the toast if so. The helper function returns null for the ignored situations.

vinceau (changes requested):

vinceau (approved):
LGTM

NikhilNarayana (approved):

PR #259, opened 2022-01-06
fix: multiple downloads and installs causing failures on macOS

NikhilNarayana:
plus some other safety changes.

### DIFF
--- a/src/dolphin/downloadDolphin.ts
+++ b/src/dolphin/downloadDolphin.ts
@@ -1,5 +1,6 @@
 import AdmZip from "adm-zip";
 import { ChildProcessWithoutNullStreams, spawn, spawnSync } from "child_process";
+import { isLinux } from "common/constants";
 import { app, BrowserWindow } from "electron";
 import { download } from "electron-dl";
 import extractDmg from "extract-dmg";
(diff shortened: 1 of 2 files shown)


src/dolphin/downloadDolphin.ts
```diff
@@ -167,7 +167,12 @@ async function installDolphin(
       }
 
       log(`Extracting to: ${dolphinPath}`);
-      await extractDmg(assetPath, dolphinPath);
+      try {
+        await extractDmg(assetPath, dolphinPath);
+      } catch (err) {
+        log("Could not extract dmg");
+        console.log(err);
+      }
```
vinceau, line 175: Why would we want to catch the error here? Wouldn't the rest of the function fail if the call to `extractDmg` failed?

NikhilNarayana: weirdly, it doesn't, the rest of the code continues and doesn't overtly fail (apart from the chmod stuff but that is in a try catch already).

Honestly it might be better to just return in that catch as a fair early.

vinceau: This is `installDolphin` function is getting kind of overly complicated.

Can you instead create 3 (or 2 if you exclude windows) new functions for installingDolphin for each individual OS?

So essentially the current `installDolphin` will at the switch statement for the OS, simply call a `installDolphinOnMac()` and `installDolphinOnLinux()` function?

Then in those OS-specific install functions, you only have to handle the optimistic path, and not need to worry about catching any errors.

The parent `installDolphin` can be the one to handle the errors if necessary.

NikhilNarayana: I left a single try catch in the Mac and Linux functions because it is likely that they wouldn't complete fully if I let the caller catch them instead.

src/dolphin/downloadDolphin.ts
```diff
@@ -145,105 +146,110 @@ async function installDolphin(
...
+    if (files) {
+      files.forEach(async (file) => {
+        if (file !== "Slippi Dolphin.app") {
+          try {
+            await fs.unlink(path.join(dolphinPath, file));
+          } catch (err) {
+            fs.removeSync(path.join(dolphinPath, file));
+          }
+        }
+      });
```
vinceau, line 209: This seems overly complicated. Aren't we just trying to find and remove files which aren't `Slippi Dolphin.app`? Maybe we could do it a bit more functionally.

```suggestion
  const files = await fs.readdir(dolphinPath);
  await Promise.all(files.filter(file => file !== "Slippi Dolphin.app").map(async (file) => {
         try {
           await fs.unlink(path.join(dolphinPath, file));
          } catch (err) {
           await fs.remove(path.join(dolphinPath, file));
          }
      }));
```

src/dolphin/downloadDolphin.ts
```diff
@@ -145,105 +146,110 @@ async function installDolphin(
...
-          log(`${dolphinAppImagePath} already exists. Deleting...`);
-          await fs.remove(dolphinAppImagePath);
-        }
-      } catch (err) {
-        log("No existing AppImage found");
-      }
+  } catch (err) {
+    log("Failed to install Dolphin...");
+    console.log(err);
+  }
```
vinceau, line 179: I don't understand why we want to catch this error here and then do nothing about it. Shouldn't the parent caller of the installDolphin function actually know if the install failed or not?

vinceau (changes requested):

src/dolphin/downloadDolphin.ts
NikhilNarayana: yeah it should probably be handled in the caller and then show a toast on failure.

src/dolphin/downloadDolphin.ts
```diff
@@ -145,100 +146,97 @@ async function installDolphin(
...
+async function installDolphinOnMac(assetPath: string, dolphinPath: string, log: (message: string) => void) {
+  const backupLocation = dolphinPath + "_old";
+  const dolphinResourcesPath = path.join(dolphinPath, "Slippi Dolphin.app", "Contents", "Resources");
+
+  const alreadyInstalled = await fs.pathExists(dolphinResourcesPath);
+  let shouldRestore = false;
+  if (alreadyInstalled) {
+    log(`${dolphinResourcesPath} already exists. Moving...`);
+    await fs.move(dolphinPath, backupLocation);
+    shouldRestore = true;
```
vinceau, line 186: Why do we track `shouldRestore`? If `alreadyInstalled` is true, then wouldn't `shouldRestore` also be guaranteed to be true?

src/dolphin/downloadDolphin.ts
```diff
@@ -145,100 +146,97 @@ async function installDolphin(
...
+
+  // move backed up User folder and user.json
+  if (alreadyInstalled) {
+    const oldUserFolder = path.join(backupLocation, "Slippi Dolphin.app", "Contents", "Resources", "User");
+    const newUserFolder = path.join(dolphinResourcesPath, "User");
+
+    if (shouldRestore) {
+      log("moving User folder...");
+      await fs.move(oldUserFolder, newUserFolder);
+    }
```
vinceau, line 216: Likewise if `alreadyInstalled` is true, then `shouldRestore` would also be guaranteed to be true. Since the only alternative is that `await fs.move` errors in which case we would never get to the rest of the function anyways. I think we can just completely remove the `shouldRestore` variable.

vinceau (changes requested):

vinceau (approved):
LGTM

PR #260, opened 2022-01-07
Migrate boilerplate to ERB

vinceau:
## Description

This PR migrates the boilerplate to use the popular and well-maintained [Electron React Boilerplate (ERB)](https://electron-react-boilerplate.js.org/). This allows us to remove the outdated and unmaintained [electron-webpack](https://webpack.electron.build/) dependency, critical for updating our Electron version. With this change we also completely remove `nodeIntegration` and instead use context isolation for better security.

## Key Changes

* Upgrades from Electron 7 to Electron 13
* Upgrades `react-router` from v5 to v6 (breaking change)
* Enables context isolation
* Uses `require.context` for stage/character image icons
* Removes references to the old `SLIPPI_USER_SERVER` in favour of using GQL endpoint for user id validation
* Removes `nodeIntegration` in renderer thread
* Node specific modules can no longer be imported from the renderer! This means no `electron-log` in renderer.
* Requires all IPC calls to be made via the API as defined in `preload.ts`
* Uses a [two package.json structure](https://www.electron.build/tutorials/two-package-structure.html). It's a bit confusing but seems to work okay.

## Todo
- [x] Fix compiled Launcher icon looking wack
- [x] Fix appData path being `slippi-launcher` instead of `Slippi Launcher`
- [x] Fix up the Electron menus
- [x] Restore the file association and URI handling
- [x] Double check app update settings
- [x] `onDragStart` is missing drag icon (fixed by downgrading Electron from v16 -> v13)
- [ ] Update CI pipelines for the new build/release locations
- [ ] Test it a lot

## Won't fix for now
* `Webpack.DefinePlugin` variables not injecting in main thread in dev mode (just don't use them in main lol)
* Sometimes CTRL+C for quitting the dev server throws some weird errors. Just ignoring it for now.

## Additional Notes

* Since `slippi-js` still relies on a lot of node builtins, we have to polyfill the node modules in order to use `slippi-js` in the renderer thread. Once slippi-js is more browser friendly we should remove all the polyfills.
* Additionally, since we have `contextIsolation` enabled now, we can safely use the `<TwitterTimelineEmbed />` react component from `react-twitter-embed`, without concerns for security.
* Auto-reloading the app when changes are made to the main thread are not supported. There might be ways to hackily create our own file watcher and restart the electron app. But honestly ceebs for now.
* This PR is only the beginning of the refactoring efforts. We'll eventually introduce dedicated services to contain functionality such as a `LoggerService` (in the absence of electron-log in renderer) and potentially send that to a centralized logging service like Sentry.


.erb/configs/webpack.config.main.prod.ts
```diff
@@ -0,0 +1,94 @@
...
+    minimizer: [
+      new TerserPlugin({
+        parallel: true,
+      }),
+    ],
+  },
+
+  plugins: [
+    new BundleAnalyzerPlugin({
+      analyzerMode: process.env.ANALYZE === "true" ? "server" : "disabled",
```
NikhilNarayana, line 64: what exactly is this doing?

NikhilNarayana (approved):
LGTM

PR #261, opened 2022-01-13
feat: update queries and mutations for new slippi backend

NikhilNarayana:


vinceau (approved):
Am just gonna trust that this works XD

PR #271, opened 2022-01-29
feat: leverage slippi backend's getLatestDolphin query

NikhilNarayana:
Swaps our current implementation of querying github in favor of the getLatestDolphin query on our backend.

The refactor of `downloadDolphin.ts` does reorder a bit of the code so the diff is a bit greater there.

### DIFF
--- a/src/dolphin/checkVersion.ts
+++ b/src/dolphin/checkVersion.ts
@@ -1,4 +1,5 @@
 import { ApolloClient, gql, HttpLink, InMemoryCache } from "@apollo/client";
+import { isDevelopment } from "common/constants";
 import { fetch } from "cross-fetch";
 import { GraphQLError } from "graphql";
 
@@ -12,7 +13,7 @@ const client = new ApolloClient({
   link: httpLink,
   cache: new InMemoryCache(),
   name: "slippi-launcher",
-  version: appVersion,
+  version: `${appVersion}${isDevelopment ? "-dev" : ""}`,
 });
 
 const getLatestDolphinQuery = gql`
--- a/src/dolphin/checkVersion.ts
+++ b/src/dolphin/checkVersion.ts
@@ -1,0 +1,62 @@
+import { ApolloClient, gql, HttpLink, InMemoryCache } from "@apollo/client";
+import { fetch } from "cross-fetch";
+import { GraphQLError } from "graphql";
+
+import { DolphinLaunchType, DolphinVersionResponse } from "./types";
+
+const httpLink = new HttpLink({ uri: process.env.SLIPPI_GRAPHQL_ENDPOINT, fetch });
+
+const appVersion = __VERSION__;
+
+const client = new ApolloClient({
+  link: httpLink,
+  cache: new InMemoryCache(),
+  name: "slippi-launcher",
+  version: appVersion,
+});
+
+const getLatestDolphinQuery = gql`
+  query GetLatestDolphin($purpose: DolphinPurpose, $includeBeta: Boolean) {
+    getLatestDolphin(purpose: $purpose, includeBeta: $includeBeta) {
+      linuxDownloadUrl
+      windowsDownloadUrl
+      macDownloadUrl
+      version
+    }
+  }
+`;
+
+const handleErrors = (errors: readonly GraphQLError[] | undefined) => {
+  if (errors) {
+    let errMsgs = "";
+    errors.forEach((err) => {
+      errMsgs += `${err.message}\n`;
+    });
+    throw new Error(errMsgs);
+  }
+};
+
+export async function fetchLatestDolphin(
+  dolphinType: DolphinLaunchType,
+  beta = false,
+): Promise<DolphinVersionResponse> {
+  const res = await client.query({
+    query: getLatestDolphinQuery,
+    fetchPolicy: "network-only",
+    variables: {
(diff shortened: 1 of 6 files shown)


src/dolphin/checkVersion.ts
```diff
@@ -0,0 +1,63 @@
...
+import { GraphQLError } from "graphql";
+
+import { DolphinLaunchType, DolphinVersionResponse } from "./types";
+
+const httpLink = new HttpLink({ uri: process.env.SLIPPI_GRAPHQL_ENDPOINT, fetch });
+
+const appVersion = __VERSION__;
+
+const client = new ApolloClient({
+  link: httpLink,
```
vinceau, line 13: Why do we need to provide a link here instead of just providing the `uri` directly? Does it not work if we don't provide it an instance to `cross-fetch`?

src/dolphin/downloadDolphin.ts
```diff
@@ -10,15 +10,27 @@ import path from "path";
...
 import { findDolphinExecutable } from "./util";
 
 function logDownloadInfo(message: string): void {
   void ipc_dolphinDownloadLogReceivedEvent.main!.trigger({ message });
 }
 
+export async function assertDolphinInstallations(): Promise<void> {
+  try {
+    await assertDolphinInstallation(DolphinLaunchType.NETPLAY, logDownloadInfo);
+    await assertDolphinInstallation(DolphinLaunchType.PLAYBACK, logDownloadInfo);
```
vinceau, line 25: Ideally we could use Promise.all so we could potentially avoid a waterfall. However, since we're sending the log to `logDownloadInfo` we might not be able to do that, since it would intersperse the log messages. Something we could keep in mind for the future though.

src/dolphin/downloadDolphin.ts
```diff
@@ -10,15 +10,27 @@ import path from "path";
...
+    await ipc_dolphinDownloadFinishedEvent.main!.trigger({ error: null });
+  } catch (err) {
+    console.error(err);
+    await ipc_dolphinDownloadFinishedEvent.main!.trigger({ error: err.message });
+  }
+  return;
```
vinceau, line 31: Can't we implicitly return? Why the explicit return here?
```suggestion
```

src/dolphin/downloadDolphin.ts
```diff
@@ -27,106 +39,63 @@ export async function assertDolphinInstallation(
     await findDolphinExecutable(type);
     log(`Found existing ${type} Dolphin executable.`);
     log(`Checking if we need to update ${type} Dolphin`);
-    const data = await getLatestReleaseData(type);
-    const latestVersion = data.tag_name;
+    const data = await fetchLatestDolphin(type);
```
vinceau, line 42:
```suggestion
    const dolphinDownloadInfo = await fetchLatestDolphin(type);
```
Can we use a better variable name than `data` here?

src/dolphin/downloadDolphin.ts
```diff
@@ -27,106 +39,63 @@ export async function assertDolphinInstallation(
...
+): Promise<void> {
+  const downloadedAsset = await downloadLatestDolphin(releaseInfo, log);
+  log(`Installing v${releaseInfo.version} ${type} Dolphin...`);
+  await installDolphin(type, downloadedAsset, log, cleanInstall);
+  log(`Finished v${releaseInfo.version} ${type} Dolphin install`);
 }
 
 async function downloadLatestDolphin(
-  type: DolphinLaunchType,
+  releaseInfo: DolphinVersionResponse,
```
vinceau, line 78:
```suggestion
  downloadUrl: string,
```
Can we pass the URL in directly here?

src/dolphin/downloadDolphin.ts
```diff
@@ -27,106 +39,63 @@ export async function assertDolphinInstallation(
...
   const exists = await fileExists(downloadLocation);
   if (!exists) {
-    log(`Downloading ${asset.browser_download_url} to ${downloadLocation}`);
+    log(`Downloading ${downloadUrl} to ${downloadLocation}`);
     const win = BrowserWindow.getFocusedWindow();
     if (win) {
-      await download(win, asset.browser_download_url, {
-        filename: asset.name,
+      await download(win, downloadUrl, {
+        filename: filename,
```
vinceau, line 94: This is just a comment, but I think it would be good if we could eventually not depend on the existence of a browser window to download a file since `electron-dl` seems to require a Browser window. I feel like there are a few cases where the app is running without a window (e.g. MacOS) or in the task bar/system tray (perhaps in the future).

vinceau (changes requested):

src/dolphin/checkVersion.ts
NikhilNarayana: Yep, we get this explicit error from ApolloClient

```
Invariant Violation:
  "fetch" has not been found globally and no fetcher has been configured. To fix this, install a fetch package (like https://www.npmjs.com/package/cross-fetch), instantiate the fetcher, and pass it into your HttpLink constructor. For example:

  import fetch from 'cross-fetch';
  import { ApolloClient, HttpLink } from '@apollo/client';
  const client = new ApolloClient({
    link: new HttpLink({ uri: '/graphql', fetch })
  });
  ```

vinceau (approved):
LGTM

PR #275, opened 2022-02-14
feat: add better handling when play key fails to fetch

JLaferri:
Previously when the user had no internet access or Slippi servers were down we would show the online activation step which was pretty confusing to users.

New behavior:
https://user-images.githubusercontent.com/1534726/153937946-e748898c-87fd-4f17-a0fc-5bd83fbbaacb.mp4

### DIFF
--- a/src/renderer/containers/Header/UserInfo.tsx
+++ b/src/renderer/containers/Header/UserInfo.tsx
@@ -11,8 +11,19 @@ export const UserInfo: React.FC<{
   uid: string;
   displayName: string | null;
   playKey: PlayKey | null;
+  isServerError: boolean | null;
   loading: boolean;
-}> = ({ uid, displayName, playKey, loading }) => {
+}> = ({ uid, displayName, playKey, isServerError, loading }) => {
+  const isErrorText = isServerError || !playKey;
+  let subtext = "";
+  if (isServerError) {
+    subtext = "Slippi server error";
+  } else if (!playKey) {
+    subtext = "Online activation required";
+  } else {
+    subtext = playKey.connectCode;
+  }
+
   return (
     <div
       css={css`
@@ -50,10 +61,10 @@ export const UserInfo: React.FC<{
             css={css`
               font-weight: bold;
               font-size: 14px;
-              color: ${playKey ? colors.purpleLight : "red"};
+              color: ${isErrorText ? "red" : colors.purpleLight};
             `}
           >
-            {playKey ? playKey.connectCode : "Online activation required"}
+            {subtext}
           </div>
         )}
       </div>
--- a/src/renderer/containers/Header/UserInfo.tsx
+++ b/src/renderer/containers/Header/UserInfo.tsx
@@ -11,12 +11,12 @@ export const UserInfo: React.FC<{
   uid: string;
   displayName: string | null;
   playKey: PlayKey | null;
-  isServerError: boolean | null;
+  serverError: boolean | null;
   loading: boolean;
-}> = ({ uid, displayName, playKey, isServerError, loading }) => {
-  const isErrorText = isServerError || !playKey;
+}> = ({ uid, displayName, playKey, serverError, loading }) => {
+  const isErrorText = serverError || !playKey;
   let subtext = "";
-  if (isServerError) {
+  if (serverError) {
     subtext = "Slippi server error";
   } else if (!playKey) {
     subtext = "Online activation required";
--- a/src/renderer/containers/Header/UserInfo.tsx
+++ b/src/renderer/containers/Header/UserInfo.tsx
@@ -14,7 +14,7 @@ export const UserInfo: React.FC<{
   serverError: boolean | null;
   loading: boolean;
 }> = ({ uid, displayName, playKey, serverError, loading }) => {
-  const isErrorText = serverError || !playKey;
+  const showError = serverError || !playKey;
   let subtext = "";
   if (serverError) {
     subtext = "Slippi server error";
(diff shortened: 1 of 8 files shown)


src/renderer/containers/Header/UserInfo.tsx
```diff
@@ -11,8 +11,19 @@ export const UserInfo: React.FC<{
   uid: string;
   displayName: string | null;
   playKey: PlayKey | null;
+  isServerError: boolean | null;
   loading: boolean;
-}> = ({ uid, displayName, playKey, loading }) => {
+}> = ({ uid, displayName, playKey, isServerError, loading }) => {
+  const isErrorText = isServerError || !playKey;
```
NikhilNarayana, line 17:
```suggestion
  const showErrorText = serverError || !playKey;
```

src/renderer/lib/hooks/useAccount.ts
```diff
@@ -34,10 +36,10 @@ export const useAccount = create(
 
         set({ loading: true });
         await fetchPlayKey()
-          .then((playKey) => set({ playKey }))
+          .then((playKey) => set({ playKey, isServerError: false }))
           .catch((err) => {
             console.warn("Error fetching play key: ", err);
-            set({ playKey: null });
+            set({ playKey: null, isServerError: true });
```
NikhilNarayana, line 42: Ok so we actually need to think of another way to do this. A user with a brand new account triggers this bit of code and thus they are never able to set their connect code.

NikhilNarayana (changes requested):
drop the `is` prefix on variables, nothing else in the launcher does it that way

src/renderer/lib/hooks/useAccount.ts
JLaferri: Huh interesting. In all likelihood I feel like an account without a play key should return a successful result with an empty play key, not an error. I can take a look tomorrow too.

JLaferri: Addressed by https://github.com/project-slippi/slippi-launcher/pull/275/commits/ac47b418e12d4fc6cbacb68d6982f2ed3dfae9f4

The `connectCode` and `private` objects on `getUser` were null for a new user.

src/renderer/lib/slippiBackend.ts
```diff
@@ -100,12 +100,20 @@ export async function fetchPlayKey(): Promise<PlayKey> {
...
   return {
     uid: user.uid,
-    connectCode: res.data.getUser.connectCode.code,
-    playKey: res.data.getUser.private.playKey,
-    displayName: res.data.getUser.displayName,
-    latestVersion: res.data.getLatestDolphin.version,
+    connectCode: connectCode,
+    playKey: playKey,
+    displayName: res.data.getUser?.displayName,
+    latestVersion: res.data.getLatestDolphin?.version,
```
vinceau, line 116:
```suggestion
    connectCode,
    playKey,
    displayName,
    latestVersion,
```

This repo uses object shorthand notation so we should stick with that.

vinceau (changes requested):

src/renderer/lib/slippiBackend.ts
```diff
@@ -100,12 +100,20 @@ export async function fetchPlayKey(): Promise<PlayKey> {
 
   handleErrors(res.errors);
 
+  const connectCode = res.data.getUser?.connectCode?.code;
+  const playKey = res.data.getUser?.private?.playKey;
+  if (!connectCode || !playKey) {
```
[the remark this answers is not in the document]
JLaferri: I agree with this from a type perspective, but not a functionality perspective. If we fail to get the version, we shouldn't be telling the user that they need to initialize their connect code.

I'd rather add null as a possible type for latest version and handle that correctly in the code.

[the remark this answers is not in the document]
JLaferri: If an optional chain like that fails, does the value become null or undefined? Do ts types differentiate between the two?

src/renderer/lib/slippiBackend.ts
JLaferri: Addressed by https://github.com/project-slippi/slippi-launcher/pull/275/commits/fdabb249215b917821f9657d80dd191b3831b81d

src/renderer/lib/slippiBackend.ts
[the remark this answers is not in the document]
JLaferri: Addressed by https://github.com/project-slippi/slippi-launcher/pull/275/commits/fdabb249215b917821f9657d80dd191b3831b81d

vinceau (approved):
Not a big fan of how type-unsafe the GQL endpoints are but maybe we can fix that later. Also we should probably eventually fix the prop drilling that's happening in the `UserInfo` component but won't block the PR for it.

JLaferri:
> Not a big fan of how type-unsafe the GQL endpoints are but maybe we can fix that later. Also we should probably eventually fix the prop drilling that's happening in the `UserInfo` component but won't block the PR for it.

GQL is never type unsafe by definition.

src/renderer/lib/hooks/useAccount.ts
```diff
@@ -11,6 +11,7 @@ export const useAccount = create(
       user: null as firebase.User | null,
       loading: false,
       playKey: null as PlayKey | null,
+      serverError: false as boolean | null,
```
NikhilNarayana, line 14: I don't see serverError ever being set to null or being set ambiguously, can we remove that extra type?

NikhilNarayana (changes requested):

src/renderer/lib/hooks/useAccount.ts
JLaferri: Probably

NikhilNarayana (approved):
LGTM

PR #278, opened 2022-02-18
Make time utils not dependent on fs

vinceau:
This PR makes the common time utils not dependent on `fs`.

The functions imported from common should work in both node and browser environments. However since the time utils import from `fs` it breaks that rule. Thus this PR moves the fs related time function into the only caller `loadFile.ts` and moves the remaining time functions into the `renderer/lib` folder.

In addition, it removes the call to `statSync` in favour of using the stat promise.


NikhilNarayana (approved):
LGTM

PR #280, opened 2022-02-23
feat: handle generating folder tree in main thread

vinceau:
This PR moves the folder tree generation logic into the main thread (node environment).

One of the last remaining blockers for the ERB migration is that there are still `fs` functions being used in the renderer thread. Specifically the generation of the folder tree structure in the replay browser was still being done on the renderer side of things. This PR fixes that such that only folder collapsing logic is handled in the renderer.


NikhilNarayana (approved):
LGTM

PR #281, opened 2022-02-26
fix: showing invalid toast immediately on drop

NikhilNarayana:
adds an `Unvalidated` state for ISOs so that we don't immediately assume the ISO is invalid and show the error


vinceau (approved):
Honestly not a fan of calling it "Unvalidated" cuz it sounds confusing as heck but just gonna approve it for now.

PR #282, opened 2022-02-26
feat: overwrite play key file if they don't match

NikhilNarayana:
in the future we want to be a bit more tolerant about slippi being in a partially down state. to handle matchmaking without the DB or firebase we need to use the details from the user.json so we should keep the file up to date.


vinceau (approved):
LGTM

PR #283, opened 2022-02-27
refactor: switch case the error codes and add some new ones

NikhilNarayana:
I'm on the fence about how I rewrote this. maybe nested switch statements are bad design, but it does make it easier to add new error codes.

### DIFF
--- a/src/renderer/lib/hooks/useAppListeners.ts
+++ b/src/renderer/lib/hooks/useAppListeners.ts
@@ -32,6 +32,7 @@ import { useAppInitialization, useAppStore } from "@/lib/hooks/useApp";
 import { useConsole } from "@/lib/hooks/useConsole";
 import { useReplays } from "@/lib/hooks/useReplays";
 
+import { handleDolphinExitCode } from "../utils";
 import { useAccount } from "./useAccount";
 import { useBroadcast } from "./useBroadcast";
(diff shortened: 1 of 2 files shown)


src/renderer/lib/hooks/useAppListeners.ts
```diff
@@ -252,17 +251,32 @@ const handleDolphinExitCode = (exitCode: number | null) => {
...
+        case 0xc0000409:
+        case 0xc000007b: {
+          return "Required DLLs for launching Dolphin are missing. Check the Help section in the settings page to fix this issue.";
+        }
+        case 0xc0000005: {
+          return "Install the latest Windows update available and then restart your computer.";
+        }
+      }
+      break;
+    }
```
vinceau, line 271: Yeah nah. Move this to a separate function like `handleWindowsErrorCode(code: number): string | null` then you can have your switch statement in there. Liekwise with the linux codes.

vinceau (changes requested):
Nested switches are 🤮🤮

src/renderer/lib/hooks/useAppListeners.ts
vinceau: Also while you're at it, move those error handling functions and the `handleDolphinExitCode` function outside of the `useAppListeners` hook. We don't really need to define them inside the main hook.

vinceau (approved):
LGTM

PR #290, opened 2022-03-07
feat: add auth service and mock auth service

vinceau:
This PR introduces the concept of a service in the renderer thread which defines an interface of methods which can be called. This service is instantiated before even the React DOM tree render and provides the services to the children using `React.useContext`.

We also introduce the ability to mock services by passing in `--env mode=mock` into the command line. This has been bound to `yarn run dev`. When this flag is enabled, the mock services will be used and provided instead of the prod services. The benefits of this is plenty:

1. This allows us to more easily develop without needing to talk to a proper backend system. 
2. This also makes it easier for other developers to join the project without needing access to a `.env` file.
3. It makes it possible in the future to compile the entire renderer app as a webpage, without any electron at all.

We utilise webpack's `NormalModuleReplacementPlugin` to swap out the modules at build time so as to not bundle any sort of mock services in production.

We distinguish between the two commands:
1. `yarn run start` (runs the development server but uses real prod services)
4. `yarn run dev` (runs the development server but uses mock services)

I'm open to swapping these around if desired since I know people are used to running `yarn run dev`, but IMO it makes more sense for the `dev` command to use the mock services rather than the other way round.

For now we only introduce an `AuthService` for handling firebase authentication, but we will in a later PR also introduce a new `SlippiBackendService` as well as other services in the future.


src/renderer/services/authService/mockAuthClient.ts
```diff
@@ -0,0 +1,60 @@
...
+
+import type { AuthService, AuthUser } from "./types";
+
+export class MockAuthClient implements AuthService {
+  private _currentUser: AuthUser | null = null;
+  private _userSubject = new Subject<AuthUser | null>();
+  private _onAuthStateChanged = multicast(this._userSubject);
+
+  public async init(): Promise<AuthUser | null> {
+    await delay(3000);
```
NikhilNarayana, line 13: is it a necessity to delay on init?

vinceau: Not necessary, but it's good for testing/simulating loading states etc.

src/renderer/services/authService/types.ts
```diff
@@ -0,0 +1,14 @@
+export type AuthUser = {
+  uid: string;
+  displayName: string;
+};
+
+export interface AuthService {
```
NikhilNarayana, line 6: Need to implement `getIdToken`.

NikhilNarayana (changes requested):
code looks fine overall, but we need to cover all the functions we use from firebase in the mock.

src/renderer/services/authService/types.ts
vinceau: I'll add that with an incoming PR for the `SlippiBackendService`.

vinceau: Should be resolved in d94ff43

NikhilNarayana (approved):
LGTM

PR #292, opened 2022-03-09
feat: handle auth, errors, and retry with links

NikhilNarayana:
retryLink is fairly straightforward so I won't explain it.

errorLink gives us some more context about failures, but throwing in that function causes really weird experiences. I can look into it further, but I'm not sure if we will gain much by throwing in there.

`setContext` is the interesting link, it essentially replaces the old `_refreshAuthToken` by handling the authn at the time of request rather than manually before a request.
Docs (it is short, I promise): https://www.apollographql.com/docs/react/api/link/apollo-link-context/

We can then combine all these links together and the apollo client will use them when necessary.

### DIFF
--- a/src/dolphin/install/fetchLatestVersion.ts
+++ b/src/dolphin/install/fetchLatestVersion.ts
@@ -1,6 +1,9 @@
-import { ApolloClient, gql, HttpLink, InMemoryCache } from "@apollo/client";
+import { ApolloClient, ApolloLink, gql, HttpLink, InMemoryCache } from "@apollo/client";
+import { onError } from "@apollo/client/link/error";
+import { RetryLink } from "@apollo/client/link/retry";
 import { appVersion } from "@common/constants";
 import { fetch } from "cross-fetch";
+import electronLog from "electron-log";
 import type { GraphQLError } from "graphql";
 
 import type { DolphinLaunchType } from "../types";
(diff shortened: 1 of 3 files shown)


src/renderer/services/slippi/slippi.service.ts
```diff
@@ -25,45 +29,54 @@ const handleErrors = (errors: readonly GraphQLError[] | undefined) => {
...
   ) {
-    this.httpLink = new HttpLink({ uri: SLIPPI_BACKEND_URL });
-    this.client = new ApolloClient({
-      link: this.httpLink,
-      cache: new InMemoryCache(),
-      name: "slippi-launcher",
-      version: clientVersion,
-    });
-  }
+    this.authService = authService;
```
vinceau, line 39:
```suggestion
```
We don't need this. Giving an accessibility modifier on a constructor param already assigns it to be accessible from within the class. i.e. `private authService: AuthService` already implies `this.authService = authService`.

src/renderer/services/slippi/slippi.service.ts
```diff
@@ -25,45 +29,54 @@ const handleErrors = (errors: readonly GraphQLError[] | undefined) => {
...
+      };
+    });
+    const retryLink = new RetryLink({
+      delay: {
+        initial: 300,
+        max: Infinity,
+        jitter: true,
+      },
+      attempts: {
+        max: 3,
```
vinceau, line 58: How come this one retries 3 times but the other one retries 5 times? Should we make it consistent?

vinceau (changes requested):

src/dolphin/install/fetchLatestVersion.ts
```diff
@@ -14,11 +17,36 @@ export type DolphinVersionResponse = {
...
+const httpLink = new HttpLink({ uri: process.env.SLIPPI_GRAPHQL_ENDPOINT, fetch });
+const retryLink = new RetryLink({
+  delay: {
+    initial: 300,
+    max: Infinity,
+    jitter: true,
+  },
+  attempts: {
+    max: 5,
+    retryIf: (error, _operation) => !!error,
```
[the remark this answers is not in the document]
vinceau: Actually can we also make this `Boolean(error)`? I prefer it over `!!`.

src/renderer/services/slippi/slippi.service.ts
NikhilNarayana: whoops, meant to make them both 3

src/renderer/services/slippi/slippi.service.ts
```diff
@@ -25,45 +29,53 @@ const handleErrors = (errors: readonly GraphQLError[] | undefined) => {
...
-      link: this.httpLink,
-      cache: new InMemoryCache(),
-      name: "slippi-launcher",
-      version: clientVersion,
-    });
-  }
+    const httpLink = new HttpLink({ uri: SLIPPI_BACKEND_URL });
+    const authLink = setContext(async () => {
+      // The firebase ID token expires after 1 hour so we will update the header on all actions
+      const token = await this.authService.getUserToken();
```
vinceau, line 42: Since the constructor is now getting pretty complicated, can you put this client creation logic into a private function such that our constructor can simply be something like:
```
this.client = this._initApolloClient(clientVersion);
```

vinceau (changes requested):

vinceau (approved):
LGTM

PR #296, opened 2022-03-19
Upgrade firebase from v8 to v9

vinceau:
### Description

This PR upgrades the `firebase` dependency from version 8 to version 9, allowing us to take advantage of tree-shaking, reducing the renderer bundle size by around 730KB. 

The `firebase` dependency seems to have an issue with ERB's postinstall script, so we just ignore it in the `renderer.dev.dll` webpack config. It seems to work fine.

#### Before
image

#### After

image


NikhilNarayana (approved):
I tried to see if I could figure out the `Package path . is not exported` issue but nothing worked. at least nothing seems to be broken.

PR #299, opened 2022-03-21
Replace react-toast-notifications with notistack

vinceau:
This PR replaces the usage of the no-longer-maintained [`react-toast-notifications`](https://www.npmjs.com/package/react-toast-notifications) with the MUI endorsed `notistack`. We need to use the v3.alpha version of [`notistack`](https://github.com/iamhosseindhv/notistack/tree/alpha) since v2 doesn't have an easy way of overwriting the toast style. Honestly not too sure if using an alpha library is better or worse than an unmaintained library. ¯\\\_(ツ)\_/¯

We also introduce our own `useToasts` hook to make future toast library changes simpler. Future toast changes should only need to update the `useToasts.ts` and `ToastProvider.tsx` files instead of all files that emit toasts.

### DIFF
--- a/src/renderer/lib/hooks/useToasts.ts
+++ b/src/renderer/lib/hooks/useToasts.ts
@@ -1,0 +1,10 @@
+import { useSnackbar } from "notistack";
+
+export const useToasts = () => {
+  const { enqueueSnackbar } = useSnackbar();
+  return {
+    showError: (message: string) => enqueueSnackbar(message, { variant: "error" }),
+    showSuccess: (message: string) => enqueueSnackbar(message, { variant: "success" }),
+    showWarning: (message: string) => enqueueSnackbar(message, { variant: "warning" }),
+  };
+};
--- a/package.json
+++ b/package.json
@@ -167,7 +167,6 @@
     "react-query": "^3.12.0",
     "react-router-dom": "^6.2.1",
     "react-timeago": "^5.2.0",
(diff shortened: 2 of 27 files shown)


src/renderer/lib/hooks/useToasts.ts
```diff
@@ -0,0 +1,10 @@
+import { useSnackbar } from "notistack";
+
+export const useToasts = () => {
+  const { enqueueSnackbar } = useSnackbar();
+  return {
+    showError: (message: string) => enqueueSnackbar(message, { variant: "error" }),
```
NikhilNarayana, line 6:
```suggestion
    showError: (message: string, options?: OptionsObject) => enqueueSnackbar(message, { ...options, variant: "error" }),
```
in case we want to pass more options from where ever these get called

OptionsObject comes from notistack

NikhilNarayana (approved):
LGTM in terms of existing notifs that I could easily test

src/renderer/lib/hooks/useToasts.ts
vinceau: I considered this too. However I think for the time being having one consistent interface for emitting notifs means that all the notifs will exude the same consistent behaviour, rather than individual notifs behaving different. It also means that if we ever change toast libraries again, we won't have to modify all the individual files that emit toasts since if they pass in an `options` object of type `OptionsObject`, they would need to also be updated to whatever the new library supports.

PR #307, opened 2022-04-11
Move dolphin installation code into its own folder

vinceau:
This PR does the following:

* moves all the dolphin installation code into its own folder for better organisation
* introduces a `DolphinInstallation` class which contains all the commonly accessed functions for interacting with an installation
* lazy-loads the per-OS installation instructions to reduce unused code
* moves all manager installation to main to minimise cross-module imports and global variable access


src/dolphin/install/installation.ts
```diff
@@ -14,59 +14,42 @@ import { fetchLatestVersion } from "./fetchLatestVersion";
...
       }
       case "darwin": {
-        sysPath = path.join(dolphinPath, "Slippi Dolphin.app", "Contents", "Resources", "Sys");
-        break;
+        return path.join(dolphinPath, "Slippi Dolphin.app", "Contents", "Resources", "Sys");
       }
       case "linux": {
-        sysPath = path.join(app.getPath("userData"), type, "Sys");
-        break;
+        return path.join(app.getPath("userData"), type, "Sys");
```
NikhilNarayana, line 50: iirc this is just the dolphin path at this point since we removed the ability to set the dolphin folder path.

NikhilNarayana (approved):
nice abstraction. made one minor note, not worth blocking this PR for

PR #309, opened 2022-04-28
Remove electron-dl dependecy

vinceau:
This PR removes the `electron-dl` dependency in favour of just downloading files manually in node. We do this because:

* `electron-dl` cannot handle multiple downloads simultaneously, despite saying otherwise
* It literally just dies when you try to download multiple things at the same time
* Has garbo documentation
* Relies on `BrowserWindow` garbo

By just relying on node, we can actually download multiple files at the same time and remove a dependency too. The only downside is we lose the flashy green downloading indicator thing on OSes. Not a big deal IMO since Netplay download and install progress will be coming soon anyways.

To replace the `electron-dl` download function, this PR extends our `download` function with the following features:
* ensuring destination folder exists
* support for tracking progress
* clean up any incomplete downloads
* lazy-load the `http` and `https` package


src/utils/download.ts
```diff
@@ -0,0 +1,79 @@
...
+          void download(nextUrl, dest, onProgress).then(() => resolve());
+          break;
+        }
+        default:
+          reject(new Error(`Download request failed, response status: ${res.statusCode} ${res.statusMessage}`));
+          break;
+      }
+    });
+
+    request.setTimeout(TIMEOUT, () => {
```
NikhilNarayana, line 74: we should have a mechanism to delete any created files if we timeout because they will be partial and thus cause future downloads to fail. I tested the scenario by artificially limiting incoming bandwidth on my machine to 100 kbps.

NikhilNarayana (changes requested):

NikhilNarayana (approved):

PR #311, opened 2022-04-29
Add DolphinService

### LINKED ISSUE #184: check for dolphin updates in the background
Users have to wait about 2 seconds every time they launch the app just for both Dolphins to be checked for updates, but updates are not very frequent. We should move the update check to the background and give users a notification telling them that there is an update available. Updates should happen on click of the notification or when the app is next launched.

vinceau:
This PR introduces a DolphinService for interacting with Dolphin features in the front-end. We also include a mock DolphinService which can simulate the Netplay Dolpin installing state.

We introduce a Dolphin installation state for both netplay and playback such that the Slippi Launcher itself can still be used while the Dolphin installation is happening. This should allow people to spend less time in the loading screen while things get installed, and only the functionality the requires Dolphin will be disabled during the installation state.

Fixes #184


NikhilNarayana (approved):
I was not ready to be pouring over this PR this weekend but great work, this is a much better experience.

PR #314, opened 2022-05-22
fix: dolphin never getting to ready state if offline

NikhilNarayana:
Users should still be able to access Dolphin related features as long as the executable is available.

### DIFF
--- a/src/dolphin/install/installation.ts
+++ b/src/dolphin/install/installation.ts
@@ -102,8 +102,8 @@ export class DolphinInstallation {
         return;
       }
 
-      const latestVersion = dolphinDownloadInfo?.version ?? "9.9.9";
-      const isOutdated = await this._isOutOfDate(latestVersion);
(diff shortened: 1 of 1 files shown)


vinceau (changes requested):

src/dolphin/install/installation.ts
```diff
@@ -141,7 +141,7 @@ export class DolphinInstallation {
         dolphinDownloadInfo = await fetchLatestVersion(type);
       } catch (err) {
         log.error(`Failed to fetch latest Dolphin version: ${err}`);
-        return;
+        throw new Error(`Failed to fetch latest Dolphin version`);
```
vinceau, line 144: I'm confused. Why even bother surrounding it in try-catch if we're just gonna throw. You could probably just do away with the try-catch block.

vinceau (changes requested):

Issue #132, opened 2021-03-16
show patch notes after an update


NikhilNarayana:
Patch notes are already listed in the news section (#129), but adding a modal to bring the notes front and center after an update is still important so that people don't accidentally miss what changed in an update.
- [ ] Display a modal that has all the patch notes for each update
  * This may require showing multiple modals or combining the patch notes.

Dgarc359:
I'm taking a stab at this

vinceau:
Not really needed anymore.

vinceau:
Not really needed anymore.

PR #315, opened 2022-05-24
feat: Patch notes modal

Dgarc359:
In reference to:
https://github.com/project-slippi/slippi-launcher/issues/132

I am re-using the logic for the news we are already getting, I spent some time trying to work with the IPC and re-use some of the github release logic, but this is what I could whip up in the quickest fashion. If there's better ways to do this I would love feedback. I was looking at the ipc but don't have much experience working with it.

### DIFF
--- a/src/renderer/containers/Header/PatchNotesDialog.tsx
+++ b/src/renderer/containers/Header/PatchNotesDialog.tsx
@@ -1,0 +1,85 @@
+import { css } from "@emotion/react";
+import { Button } from "@mui/material";
+import Dialog from "@mui/material/Dialog";
+import DialogContent from "@mui/material/DialogContent";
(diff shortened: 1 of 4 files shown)


vinceau:
This is a nice attempt! Unfortunately it doesn't really provide much value since the news feed is basically also one click away. Also issue #132 is really old and honestly isn't really necessary anymore. Keep your eye out on the `#find-work` channel to see if there are more relevant tasks that come up :)

vinceau:
This is a nice attempt! Unfortunately it doesn't really provide much value since the news feed is basically also one click away. Also issue #132 is really old and honestly isn't really necessary anymore. Keep your eye out on the `#find-work` channel to see if there are more relevant tasks that come up :)

PR #320, opened 2022-06-06
feat: pass consoleNick or broadcaster connect code to comm file when spectating

NikhilNarayana:
dolphin will take the name and write to the window to make it easier for streamers to identify which dolphin instance is which.

### DIFF
--- a/src/broadcast/spectateManager.ts
+++ b/src/broadcast/spectateManager.ts
@@ -29,8 +29,8 @@ const generatePlaybackId = (broadcastId: string) => `spectate-${broadcastId}`;
  * Dealing with dolphin related details should be handled elsewhere.
  */
 export class SpectateManager extends EventEmitter {
-  private broadcastInfo: Record<string, BroadcastInfo> = {};
-  private broadcastList: Record<string, BroadcasterItem> = {};
+  private openBroadcasts: Record<string, BroadcastInfo> = {};
+  private availableBroadcasts: Record<string, BroadcasterItem> = {};
   private wsConnection: connection | null = null;
 
   constructor() {
@@ -45,7 +45,7 @@ export class SpectateManager extends EventEmitter {
     const events = obj.events ?? [];
     const broadcastId: string = obj.broadcastId;
 
-    const broadcastInfo = this.broadcastInfo[broadcastId];
+    const broadcastInfo = this.openBroadcasts[broadcastId];
     if (!broadcastInfo) {
       // We've stopped watching this broadcast already
       return;
@@ -111,7 +111,7 @@ export class SpectateManager extends EventEmitter {
         this.wsConnection = connection;
 
         // Reconnect to all the broadcasts that we were already watching
-        Object.entries(this.broadcastInfo).forEach(([broadcastId, info]) => {
+        Object.entries(this.openBroadcasts).forEach(([broadcastId, info]) => {
           const watchMsg: { type: string; broadcastId: string; startCursor?: string } = {
             type: "watch-broadcast",
             broadcastId,
@@ -159,8 +159,10 @@ export class SpectateManager extends EventEmitter {
             case "list-broadcasts-resp": {
               const broadcasts: BroadcasterItem[] = obj.broadcasts ?? [];
               this.emit(SpectateEvent.BROADCAST_LIST_UPDATE, broadcasts);
-              this.broadcastList = {};
-              broadcasts.forEach((broadcast) => (this.broadcastList[broadcast.id] = broadcast));
+              this.availableBroadcasts = {};
+              broadcasts.forEach((broadcast) => {
+                this.availableBroadcasts[broadcast.id] = broadcast;
+              });
               break;
             }
             case "events": {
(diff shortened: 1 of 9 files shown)


src/console/types.ts
```diff
@@ -20,6 +20,7 @@ export interface MirrorConfig {
   enableRelay: boolean;
   autoSwitcherSettings?: AutoSwitcherSettings;
   useNicknameFolders: boolean;
+  consoleNick?: string;
```
vinceau, line 23:
```suggestion
  nickname?: string;
```
Can we just call this `nickname`? I actually hate that somehow we ended up with `consoleNick` somewhere in our API but there's no reason why we need to use that name here.

src/console/mirror.worker.ts
```diff
@@ -79,7 +92,12 @@ const methods: WorkerSpec = {
   getErrorObservable(): Observable<Error | string> {
     return Observable.from(errorSubject);
   },
-  getMirrorDetailsObservable(): Observable<{ playbackId: string; filePath: string; isRealtime: boolean }> {
+  getMirrorDetailsObservable(): Observable<{
+    playbackId: string;
+    filePath: string;
+    isRealtime: boolean;
+    consoleNick?: string;
```
vinceau, line 99:
```suggestion
    nickname?: string;
```

src/console/mirror.worker.ts
```diff
@@ -40,9 +50,12 @@ mirrorManager.on(MirrorEvent.ERROR, async (error: Error | string) => {
   errorSubject.next(error);
 });
 
-mirrorManager.on(MirrorEvent.NEW_FILE, async (playbackId: string, filePath: string, isRealtime: boolean) => {
-  mirrorDetailsSubject.next({ playbackId, filePath, isRealtime });
-});
+mirrorManager.on(
+  MirrorEvent.NEW_FILE,
+  async (playbackId: string, filePath: string, isRealtime: boolean, consoleNick?: string) => {
+    mirrorDetailsSubject.next({ playbackId, filePath, isRealtime, consoleNick });
```
vinceau, line 56:
```suggestion
  async (playbackId: string, filePath: string, isRealtime: boolean, nickname?: string) => {
    mirrorDetailsSubject.next({ playbackId, filePath, isRealtime, nickname });
```

src/console/mirror.worker.ts
```diff
@@ -28,7 +33,12 @@ const mirrorManager = new MirrorManager();
 
 const logSubject = new Subject<string>();
 const errorSubject = new Subject<Error | string>();
-const mirrorDetailsSubject = new Subject<{ playbackId: string; filePath: string; isRealtime: boolean }>();
+const mirrorDetailsSubject = new Subject<{
+  playbackId: string;
+  filePath: string;
+  isRealtime: boolean;
+  consoleNick?: string;
```
vinceau, line 40:
```suggestion
  nickname?: string;
```

src/console/mirror.worker.ts
```diff
@@ -18,7 +18,12 @@ interface Methods {
   dolphinClosed(playbackId: string): Promise<void>;
   getLogObservable(): Observable<string>;
   getErrorObservable(): Observable<Error | string>;
-  getMirrorDetailsObservable(): Observable<{ playbackId: string; filePath: string; isRealtime: boolean }>;
+  getMirrorDetailsObservable(): Observable<{
+    playbackId: string;
+    filePath: string;
+    isRealtime: boolean;
+    consoleNick?: string;
```
vinceau, line 25:
```suggestion
    nickname?: string;
```

src/broadcast/spectateManager.ts
```diff
@@ -30,14 +30,15 @@ const generatePlaybackId = (broadcastId: string) => `spectate-${broadcastId}`;
  */
 export class SpectateManager extends EventEmitter {
   private broadcastInfo: Record<string, BroadcastInfo> = {};
+  private broadcastList: Record<string, BroadcasterItem> = {};
```
vinceau, line 33: These two variables are too similar in name. It's really not clear what contains what. I think we should rename them. Maybe something like this?
```suggestion
  private currentlySpectating: Record<string, BroadcastInfo> = {};
  private availableBroadcasts: Record<string, BroadcasterItem> = {};
```
Not sure on the name of `currentlySpectating` but something like `activeBroadcasts` isn't that much more descriptive, cuz technically an "available" broadcast is also an active one.

src/broadcast/spectateManager.ts
```diff
@@ -158,6 +159,8 @@ export class SpectateManager extends EventEmitter {
             case "list-broadcasts-resp": {
               const broadcasts: BroadcasterItem[] = obj.broadcasts ?? [];
               this.emit(SpectateEvent.BROADCAST_LIST_UPDATE, broadcasts);
+              this.broadcastList = {};
+              broadcasts.forEach((broadcast) => (this.broadcastList[broadcast.id] = broadcast));
```
vinceau, line 163:
```suggestion
              this.broadcastList = {};
              broadcasts.forEach((broadcast) => {
                this.broadcastList[broadcast.id] = broadcast;
              });
```
Can we have this in a curly brace?

vinceau (changes requested):
Looks fine for the most part. Just some naming issues.

vinceau (approved):
LGTM

PR #321, opened 2022-06-08
Allow absolute HTTP replay path

willthornton:
The slippi:// protocol currently hard-codes part of the URL. This update
allows absolute URLs to be passed while not breaking current slippi.gg
site functionality. This will enable the protocol to be used on third
party sites that wish to serve slippi launcher links for their replay
repositories.


src/main/main.ts
```diff
@@ -223,7 +223,11 @@ const handleSlippiURIAsync = async (aUrl: string) => {
```
willthornton, line 223: path.basename on line 222 functions as expected when passed an absolute HTTP URL

src/main/main.ts
```diff
@@ -223,7 +223,12 @@ const handleSlippiURIAsync = async (aUrl: string) => {
 
       const fileAlreadyExists = await fileExists(destination);
       if (!fileAlreadyExists) {
-        const dlUrl = `https://storage.googleapis.com/slippi.appspot.com/${replayPath}`;
+        let dlUrl;
+        if (replayPath.startsWith("http")) {
+          dlUrl = replayPath;
+        } else {
+          dlUrl = `https://storage.googleapis.com/slippi.appspot.com/${replayPath}`;
+        }
```
vinceau, line 231:
```suggestion
        const dlUrl = (replayPath.startsWith("http")) ? replayPath : `https://storage.googleapis.com/slippi.appspot.com/${replayPath}`;
```
Can we maybe just use a ternary operator here?

vinceau (changes requested):
What are the security implications of this, if any? For example, using the slippi protocol to get users to download malicious files onto their system.

vinceau (approved):
After further consideration, I don't think allowing other websites to access the `slippi://` protocol is much worse than what people could already do with manually spreading of malicious SLP files. i.e. the status quo of current attack vectors remains largely unchanged even after this is merged. So I think it should be fine. We may reassess this change again in the future should the need arise.

PR #322, opened 2022-06-09
deps: swap adm-zip for node-stream-zip

NikhilNarayana:
tested on all 3 OSs without issues.


src/dolphin/install/linux.ts
```diff
@@ -1,7 +1,7 @@
 import type { DolphinLaunchType } from "@dolphin/types";
 import { findDolphinExecutable } from "@dolphin/util";
-import AdmZip from "adm-zip";
 import * as fs from "fs-extra";
+import StreamZip from "node-stream-zip";
```
vinceau, line 4:
```suggestion
import { async as AsyncStreamZip }  from "node-stream-zip";
```
Pretty sure we can just do this and remove those eslint config lines, with `new AsyncStreamZip()`.

vinceau (changes requested):

vinceau (approved):

PR #335, opened 2022-09-01
Update for OBS Websocket 5.0

NikhilNarayana:
So the autoswitcher works with OBS 28+.

Still need to test with 2 PCs.

### DIFF
--- a/src/console/autoSwitcher.ts
+++ b/src/console/autoSwitcher.ts
@@ -10,7 +10,7 @@ export class AutoSwitcher extends EventEmitter {
   private obsIP: string;
   private obsPassword?: string;
   private statusOutput: { status: boolean; timeout: NodeJS.Timeout | null };
-  private obsPairs: { scene: string; id: number; name: string }[];
+  private obsPairs: { scene: string; id: number }[];
 
   constructor(settings: AutoSwitcherSettings) {
     super();
@@ -43,13 +43,13 @@ export class AutoSwitcher extends EventEmitter {
       const sceneItemListRes = await this.obs.call("GetSceneItemList", { sceneName });
       const sources = sceneItemListRes.sceneItems;
       sources.forEach((source) => {
-        if (source.sourceName!.toString() === this.obsSourceName) {
-          this.obsPairs.push({
-            scene: sceneName,
-            id: +source.sceneItemId!.toString(),
-            name: source.sourceName!.toString(),
-          });
+        if (source.sourceName!.toString() !== this.obsSourceName) {
+          return;
         }
+        this.obsPairs.push({
+          scene: sceneName,
+          id: +source.sceneItemId!.toString(),
+        });
       });
     });
   };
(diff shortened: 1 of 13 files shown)


src/console/mirrorManager.ts
```diff
@@ -193,7 +193,7 @@ export class MirrorManager extends EventEmitter {
 
     details.connection.disconnect();
     if (details.autoSwitcher) {
-      details.autoSwitcher.disconnect();
+      void details.autoSwitcher.disconnect();
```
vinceau, line 196: Can we just make the disconnect function an async function and then use await here? We really should avoid using `void` unless we have to like in a callback or something.

vinceau (changes requested):

src/console/autoSwitcher.ts
```diff
@@ -1,62 +1,77 @@
 import { EventEmitter } from "events";
-import OBSWebSocket from "obs-websocket-js";
+import { isIPv6 } from "net";
```
vinceau, line 2: If this function exists (as well as `isIPv4`) in the built in `net` package, how come we need to install the `is-ip` package?

src/renderer/lib/validate.ts
```diff
@@ -1,28 +1,15 @@
-export const isValidIpAddress = (ip: string): boolean => {
-  if (ip === "localhost") {
-    return true;
-  }
-
-  const chunks = ip.split(".");
-  if (chunks.length !== 4) {
-    return false;
-  }
+import { isIPv4, isIPv6 } from "is-ip";
```
vinceau, line 1:
```suggestion
import { isIPv4, isIPv6 } from "net";
```
What's the difference from just importing these from the `net` package?

src/renderer/containers/Console/AddConnectionForm.tsx
```diff
@@ -187,7 +189,35 @@ export const AddConnectionForm: React.FC<AddConnectionFormProps> = ({ defaultVal
...
+                      required: enableAutoSwitcher,
+                      validate: (val) => {
+                        if (!enableAutoSwitcher) {
+                          return true;
+                        }
+                        if (!val) {
+                          return false;
+                        }
+                        const port = parseInt(val);
+                        return (port > 0 && port < 65535) || "Invalid Port";
```
vinceau, line 220:
```suggestion
                        return val && isValidPort(val) || "Invalid Port";
```
Should we have this in a separate function?

```typescript
function isValidPort(value: string | number) {
  const port = typeof value === "string" ? parseInt(value) : value;
  return port > 0 && port < 65535);
}
```

vinceau (changes requested):
Looks fine for the most part but have a few nits.

src/renderer/lib/validate.ts
NikhilNarayana: we can't use `net.isIP` from renderer. I could add IPC but that seems overkill for this.

vinceau: Oh right. We could consider exposing it over the context bridge, similar to how to expose the the electron-log functions. E.g.
```
     // inside of preload.ts
     log: log.functions,
     net: { isIPv4, isIPv6},
```
and then in the actual usage of it:
```
    const { isIPv4, isIPv6 } = window.electron.net; 
```

Although it does reduce a dependency, it also makes it harder to test logic and components since it now requires them to be injected over context bridge.

I'll defer it to you, but I'm happy to approve this change once the tests/build is passing.

vinceau (approved):
LGTM once the build/tests are passing.

PR #340, opened 2022-09-15
[WIP] [PR] Handle netplay Dolphin update via URI

JLaferri:
The idea behind this is that when Dolphin is not up to date, it shows an update button. when someone presses that update button, we'd like to update their Dolphin instead of requiring them to restart the launcher.

Currently testing this from the command line and it does seem to update the Dolphin version but it doesn't show the "Updating" UI, would probably like to add that.

### DIFF
--- a/release/app/package.json
+++ b/release/app/package.json
@@ -2,7 +2,7 @@
   "name": "slippi-launcher",
   "productName": "Slippi Launcher",
   "description": "Launch Slippi Online, browse and watch saved replays",
-  "version": "2.5.6",
+  "version": "2.5.7-beta.1",
   "main": "./dist/main/main.js",
   "author": {
     "name": "Jas Laferriere",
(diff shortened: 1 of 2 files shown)


JLaferri:
image

Another potential thing to consider is that if someone updates without having the Launcher installed they will see this. Afaik there's no great way to check if the launcher exists. Perhaps we could install another app that can be launched from the command line?

The backup option if they don't have the Launcher installed or if the URI fails (I don't know how to detect this), would be to send the person here: https://slippi.gg/downloads?update=true

zDEFz:
> The idea behind this is that when Dolphin is not up to date, it shows an update button. when someone presses that update button, we'd like to update their Dolphin instead of requiring them to restart the launcher.
> 
> Currently testing this from the command line and it does seem to update the Dolphin version but it doesn't show the "Updating" UI, would probably like to add that.

Proposal: Checksum part of the Launcher and if it differs, then invoke a function that downloads the newer files / cleans up, after SlippiLauncher has been quit. If the user wants to play with an outdated version, they'll have to update.

PR #341, opened 2022-09-22
feat: soft reset

NikhilNarayana:
Replace `Clear Cache` with `Soft Reset` and rename `Reset everything` to `Hard Reset`. Soft Reset is just a reinstall without removing user settings, apart from clearing the cache.

I did mess with the UI a bit, both of the reset options have indicators so I made the buttons wider to ensure they don't change size when the indicator shows up.

image

### DIFF
--- a/src/dolphin/manager.ts
+++ b/src/dolphin/manager.ts
@@ -151,7 +151,7 @@ export class DolphinManager {
     }
   }
 
-  public async reinstallDolphin(launchType: DolphinLaunchType, cleanInstall: boolean) {
+  public async reinstallDolphin(launchType: DolphinLaunchType, cleanInstall?: boolean) {
     switch (launchType) {
       case DolphinLaunchType.NETPLAY: {
         if (this.netplayDolphinInstance !== null) {
--- a/src/dolphin/manager.ts
+++ b/src/dolphin/manager.ts
@@ -151,7 +151,7 @@ export class DolphinManager {
     }
   }
 
-  public async reinstallDolphin(launchType: DolphinLaunchType) {
+  public async reinstallDolphin(launchType: DolphinLaunchType, cleanInstall: boolean) {
     switch (launchType) {
       case DolphinLaunchType.NETPLAY: {
         if (this.netplayDolphinInstance !== null) {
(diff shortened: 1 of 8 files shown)


src/dolphin/manager.ts
```diff
@@ -151,7 +151,7 @@ export class DolphinManager {
     }
   }
 
-  public async reinstallDolphin(launchType: DolphinLaunchType) {
+  public async reinstallDolphin(launchType: DolphinLaunchType, cleanInstall: boolean) {
```
vinceau, line 154:
```suggestion
  public async reinstallDolphin(launchType: DolphinLaunchType, cleanInstall?: boolean) {
```
nit: can we keep this optional?

src/renderer/containers/Settings/DolphinSettings.tsx
```diff
@@ -28,12 +34,11 @@ export const DolphinSettings: React.FC<{ dolphinType: DolphinLaunchType }> = ({
   );
   const [dolphinPath, setDolphinPath] = useDolphinPath(dolphinType);
   const [resetModalOpen, setResetModalOpen] = React.useState(false);
-  const [isResetting, setIsResetting] = React.useState(false);
+  const [isResetting, setIsResetting] = React.useState(ResetType.NONE);
```
vinceau, line 37:
```suggestion
  const [isResetType, setResetType] = React.useState(ResetType.NONE);
```
This variable needs to be renamed too.

vinceau (approved):
LGTM once these comments are addressed.

src/renderer/containers/Settings/DolphinSettings.tsx
```diff
@@ -19,6 +19,12 @@ import { SettingItem } from "./SettingItem";
 const { isLinux, isMac } = window.electron.common;
 const log = window.electron.log;
 
+enum ResetType {
+  NONE,
```
vinceau, line 23:
```suggestion
```
Also, I don't think we need a `NONE` reset type. Instead we could do `React.useState<ResetType | null>(null);` with a `null` initial value.

PR #349, opened 2022-11-10
[PR] Email Verification

JLaferri:


src/renderer/services/auth/types.ts
```diff
@@ -10,6 +12,8 @@ export interface AuthService {
   init(): Promise<AuthUser | null>;
   login(args: { email: string; password: string }): Promise<AuthUser | null>;
   logout(): Promise<void>;
+  reloadUser(): Promise<AuthUser | null>;
+  verifyEmail(): Promise<void>;
```
vinceau, line 16:
```suggestion
  refreshUser(): Promise<void>;
  sendVerificationEmail(): Promise<void>;
```
Can we rename these methods? I feel like these method names are a bit more indicative of what the function should do.

Rather than returning the `AuthUser | null` object in `refreshUser` I would prefer us just notify the listeners of the new user object using the observable subject. See the relevant comment in `auth.service.ts`.

src/renderer/services/auth/auth.service.ts
```diff
@@ -107,9 +117,37 @@ class AuthClient implements AuthService {
...
+  public async reloadUser(): Promise<AuthUser | null> {
+    const auth = getAuth();
+    const user = auth.currentUser;
+    if (!user) {
+      throw new Error("User is not logged in.");
+    }
+
+    await user.reload();
+    return auth.currentUser ? this._mapFirebaseUserToAuthUser(auth.currentUser) : null;
+  }
```
vinceau, line 149:
```suggestion
  public async refreshUser(): Promise<void> {
    const auth = getAuth();
    const user = auth.currentUser;
    if (!user) {
      throw new Error("User is not logged in.");
    }

    await user.reload();
    // Notify listeners of the new user object
    this._userSubject.next(this.getCurrentUser());
  }
```

src/renderer/services/auth/auth.service.ts
```diff
@@ -107,9 +117,37 @@ class AuthClient implements AuthService {
     if (!user) {
       return null;
     }
+
     return this._mapFirebaseUserToAuthUser(user);
   }
 
+  public async verifyEmail() {
```
vinceau, line 124:
```suggestion
  public async sendVerificationEmail() {
```

src/renderer/containers/QuickStart/VerifyEmailStep.tsx
```diff
@@ -0,0 +1,138 @@
...
+
+export const VerifyEmailStep: React.FC = () => {
+  const { authService } = useServices();
+  const setServerError = useAccount((store) => store.setServerError);
+  const setUser = useAccount((store) => store.setUser);
+  const user = useAccount((store) => store.user);
+  const emailVerificationSent = useAccount((store) => store.emailVerificationSent);
+  const setEmailVerificationSent = useAccount((store) => store.setEmailVerificationSent);
+
+  const profileUrl = "https://slippi.gg/profile";
```
vinceau, line 77:
```suggestion
  const profileUrl = "https://slippi.gg/profile";
```
Can we either declare this constant outside of the React component, or put the URL in `src/common/constants.ts`?

src/renderer/containers/QuickStart/VerifyEmailStep.tsx
```diff
@@ -0,0 +1,138 @@
...
+    <div css={classes.confirmationContainer}>
+      <CheckCircleOutlineIcon />
+      Email verified
+    </div>
+  );
+
+  return (
+    <Box display="flex" flexDirection="column" flexGrow="1">
+      <Container>
+        <QuickStartHeader>Verify Email</QuickStartHeader>
```
vinceau, line 128:
```suggestion
        <QuickStartHeader>Verify your email</QuickStartHeader>
```
All the other headings use sentence casing so we should probably be consistent.

src/renderer/containers/QuickStart/VerifyEmailStep.tsx
```diff
@@ -0,0 +1,138 @@
...
+    </>
+  );
+
+  const postVerification = (
+    <div css={classes.confirmationContainer}>
+      <CheckCircleOutlineIcon />
+      Email verified
+    </div>
+  );
+
```
vinceau, line 124:
```suggestion
  if (!user) {
    return null; // Or a message of some sort
  }
```
Can we handle the null user case earlier to avoid having to do optional chaining below?

src/renderer/containers/QuickStart/VerifyEmailStep.tsx
```diff
@@ -0,0 +1,138 @@
...
+  const profileUrl = "https://slippi.gg/profile";
+
+  const handleCheckVerification = async () => {
+    try {
+      const newUser = await authService.reloadUser();
+      setUser(newUser);
+    } catch (err: any) {
+      setServerError(err.message);
+    }
+  };
```
vinceau, line 86:
```suggestion
  const handleCheckVerification = () => {
    authService.refreshUser().catch(err => {
      setServerError(err.message);
    });
  };
```
If we let the `refreshUser` method handle the notifying of user object changes (which already gets forwarded to the store), we don't have to handle user setting at all in this component.

src/renderer/containers/QuickStart/VerifyEmailStep.tsx
```diff
@@ -0,0 +1,138 @@
...
+      if (user && !user.emailVerified && !emailVerificationSent) {
+        try {
+          await authService.verifyEmail();
+          setEmailVerificationSent(true);
+        } catch (err: any) {
+          setServerError(err.message);
+        }
+      }
+    })();
+  }, [emailVerificationSent, setEmailVerificationSent, setServerError, user, authService]);
```
vinceau, line 99:
```suggestion
  useEffect(() => {
      const sendVerificationEmail = async () => {
        try {
          await authService.verifyEmail();
          setEmailVerificationSent(true);
        } catch (err: any) {
          setServerError(err.message);
     };
     if (user && !user.emailVerified && !emailVerificationSent) {
       void sendVerificationEmail();
     }
  }, [emailVerificationSent, setEmailVerificationSent, setServerError, user, authService]);
```
I think it's a little cleaner if we make the async handling a named function.

vinceau (changes requested):
Seems fine for the most part, just some mostly code style changes.

vinceau (approved):
LGTM but still have that one nit about checking for a null `user` in `VerifyEmailStep.tsx`.

JLaferri:
> LGTM but still have that one nit about checking for a null `user` in `VerifyEmailStep.tsx`.

Yeah I was looking at this one now but idk... it feels kind of weird to me to just return nothing in the case there's no User? Honestly I'm not even sure if it's possible to be in that state at all though.

vinceau:
> > LGTM but still have that one nit about checking for a null `user` in `VerifyEmailStep.tsx`.
> 
> Yeah I was looking at this one now but idk... it feels kind of weird to me to just return nothing in the case there's no User? Honestly I'm not even sure if it's possible to be in that state at all though.

I think it should be fine in the sense that, in order to verify an email, we would need a user. So the context of verifying an email with a null user, should mean it's okay to just return null. But you're right it shouldn't be a case that should happen but I think it's still good to be defensive.

JLaferri:
> > > LGTM but still have that one nit about checking for a null `user` in `VerifyEmailStep.tsx`.
> > 
> > 
> > Yeah I was looking at this one now but idk... it feels kind of weird to me to just return nothing in the case there's no User? Honestly I'm not even sure if it's possible to be in that state at all though.
> 
> I think it should be fine in the sense that, in order to verify an email, we would need a user. So the context of verifying an email with a null user, should mean it's okay to just return null. But you're right it shouldn't be a case that should happen but I think it's still good to be defensive.

Yeah but this is UI, displaying nothing is not very helpful. I probably should display an error I guess, but that requires me to restructure a few things. I'll try something and see if you're okay with it.

Issue #336, opened 2022-09-02
Display Dolphin versions in launcher


NikhilNarayana:
Would be helpful for easily telling what is installed rather than having to boot dolphin to check.

beaudrychase:
Hello
I was thinking of picking this up as my first issue in this project, and I wanted to get an opinion on where we want this info shown.
I was thinking it would be good to add the info to the sidebar in settings.
![Capture](https://user-images.githubusercontent.com/9560811/199562335-f890ea82-f516-460f-a459-a310fdbbaca1.PNG)

zaderrr:
Had a brief go and put the version at the top.
![Capture](https://user-images.githubusercontent.com/56570635/201894828-f072796f-1c3e-40be-9fcd-69b39784c9d4.PNG)
Not sure where would be better?

vinceau:
Fixed in #351

PR #351, opened 2022-11-16
Show dolphin version display in launcher

zaderrr:
This addresses #336.

Capture
Displays dolphin version underneath the respective type (Netplay & Playback)

The version gets grabbed after it's confirmed that a Dolphin.exe exists

If there is no Dolphin exec, display "Version: Not Found"
if the version cannot be found, display "Version: Unkown"

Only tested on Windows


src/renderer/lib/dolphin/useDolphinActions.ts
```diff
@@ -51,6 +51,22 @@ export const useDolphinActions = (dolphinService: DolphinService) => {
...
+    async (dolphinType: DolphinLaunchType): Promise<string> => {
+      try {
+        const version = await dolphinService.getDolphinVersion(dolphinType);
+        const rtn: string = version["result"]["version"];
+        if (rtn === "") {
+          return "Unkown";
+        }
+        return rtn;
+      } catch (err) {
+        return "Unkown";
```
vinceau, line 64:
```suggestion
        return "";
```
I'd rather the individual component choose to handle rendering an undefined or empty string.

vinceau (changes requested):
So for the most part these changes look fine. Could you, however, instead experiment with just sending the version back with the `DolphinDownloadCompleteEvent`? Rather than defining a whole other `getDolphinVersion` IPC method etc.

For example, extending the type to be as follows:
```typescript
// In src/dolphin/types.ts
export type DolphinDownloadCompleteEvent = {
  type: DolphinEventType.DOWNLOAD_COMPLETE;
  dolphinType: DolphinLaunchType;
  version: string;
};
```

zaderrr:
Thanks for the advice. Nuked the IPC methods and sets after installation is completed.

Will add that I can't test what it looks like whilst the download is in progress, I'd imagine it will show "Not Found" until it's set when the download is complete.

NikhilNarayana (changes requested):
small nits, but otherwise this is looking good. thanks

src/renderer/lib/dolphin/useDolphinStore.ts
```diff
@@ -14,6 +14,8 @@ export const useDolphinStore = create(() => ({
   netplayOpened: false,
   playbackOpened: false,
   netplayDownloadProgress: null as { current: number; total: number } | null,
+  netplayDolphinVersion: "Not Found",
+  playbackDolphinVersion: "Not Found",
```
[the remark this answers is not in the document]
zaderrr: The reason for Not Found being default is that, currently, the version only gets grabbed when dolphin is confirmed to be present from `dolphinCompleteHandler`
So, if the default was Unknown & dolphin didn't download, it would remain as unknown rather than Not found.

Think I mentioned it in another reply but also this way whilst dolphin is being downloaded it will show not found until it is downloaded I think.

Can change the defaults, but I feel like if I were to see it say "Unknown" I'd wonder why it's not recognising my dolphin rather than seeing "Not found" then I know there is no dolphin in the directory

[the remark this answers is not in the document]
zaderrr: also I tested with a regular version of dolphin, it returns an empty string, so if someone put in a normal version of dolphin it would show unkown, as it's not a recognised slippi dolphin version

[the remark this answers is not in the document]
NikhilNarayana: I'm happy with this explanation.

NikhilNarayana (approved):

src/dolphin/install/installation.ts
```diff
@@ -176,11 +176,16 @@ export class DolphinInstallation {
   }
 
   private async _isOutOfDate(latestVersion: string): Promise<boolean> {
+    const dolphinVersion = await this.getDolphinVersion();
+    return lt(dolphinVersion, latestVersion);
+  }
+
+  public async getDolphinVersion(): Promise<string> {
```
vinceau, line 183:
```suggestion
  public async getDolphinVersion(): Promise<string | null> {
```
I think we should return null instead of an empty string if we can't find the dolphin executable.

src/dolphin/install/installation.ts
```diff
@@ -176,11 +176,16 @@ export class DolphinInstallation {
...
     const dolphinPath = await this.findDolphinExecutable();
     const dolphinVersionOut = spawnSync(dolphinPath, ["--version"]).stdout.toString();
     const match = dolphinVersionOut.match(semverRegex);
     const dolphinVersion = match?.[0] ?? "";
-    return lt(dolphinVersion, latestVersion);
+    return dolphinVersion;
```
vinceau, line 188: Can we wrap this in a try catch block, returning null if there's no match, and also if an error is thrown?

src/dolphin/install/installation.ts
```diff
@@ -176,11 +176,16 @@ export class DolphinInstallation {
   }
 
   private async _isOutOfDate(latestVersion: string): Promise<boolean> {
+    const dolphinVersion = await this.getDolphinVersion();
+    return lt(dolphinVersion, latestVersion);
```
vinceau, line 180:
```suggestion
    const dolphinVersion = await this.getDolphinVersion();
    return dolphinVersion && lt(dolphinVersion, latestVersion);
```
If `dolphinVersion` is updated to be null we probably want to ensure it actually exists too.

src/dolphin/manager.ts
```diff
@@ -30,7 +30,7 @@ export class DolphinManager {
     const dolphinInstall = this.getInstallation(dolphinType);
     await dolphinInstall.validate({
       onProgress: (current, total) => this._onProgress(dolphinType, current, total),
-      onComplete: () => this._onComplete(dolphinType),
+      onComplete: async () => this._onComplete(dolphinType, await dolphinInstall.getDolphinVersion()),
```
vinceau, line 33:
```suggestion
      onComplete: () => dolphinInstall.getDolphinVersion().then(version => {
        this._onComplete(dolphinType, version); 
      }),
```

I think I would prefer the callback to be synchronous.

src/dolphin/manager.ts
```diff
@@ -211,10 +211,11 @@ export class DolphinManager {
     });
   }
 
-  private _onComplete(dolphinType: DolphinLaunchType) {
+  private _onComplete(dolphinType: DolphinLaunchType, dolphinVersion: string) {
```
vinceau, line 214:
```suggestion
  private _onComplete(dolphinType: DolphinLaunchType, dolphinVersion: string | null) {
```
I think all of these need to be updated to support null.

src/dolphin/types.ts
```diff
@@ -71,6 +71,7 @@ export type DolphinDownloadProgressEvent = {
 export type DolphinDownloadCompleteEvent = {
   type: DolphinEventType.DOWNLOAD_COMPLETE;
   dolphinType: DolphinLaunchType;
+  dolphinVersion: string;
```
vinceau, line 74:
```suggestion
  dolphinVersion: string | null;
```

src/renderer/lib/dolphin/useDolphinListeners.ts
```diff
@@ -35,8 +41,9 @@ export const useDolphinListeners = (dolphinService: DolphinService) => {
     }
   }, []);
 
-  const dolphinCompleteHandler = useCallback((event: DolphinDownloadCompleteEvent) => {
+  const dolphinCompleteHandler = useCallback(async (event: DolphinDownloadCompleteEvent) => {
```
vinceau, line 44:
```suggestion
  const dolphinCompleteHandler = useCallback((event: DolphinDownloadCompleteEvent) => {
```
this probably doesn't still need to be asynchronous

vinceau (changes requested):
Getting there!

src/dolphin/install/installation.ts
zaderrr: Not sure about wrapping in a try catch as this function only gets run if it is confirmed that a dolphin exec exists.

zaderrr:
Fixed suggested issues, but didn't wrap in a try catch block as I feel it's unnecessary as we are only running that function when a dolphin exec exists.

src/renderer/lib/dolphin/useDolphinStore.ts
```diff
@@ -14,6 +14,8 @@ export const useDolphinStore = create(() => ({
   netplayOpened: false,
   playbackOpened: false,
   netplayDownloadProgress: null as { current: number; total: number } | null,
+  netplayDolphinVersion: "Not Found" as string | null,
+  playbackDolphinVersion: "Not Found" as string | null,
```
vinceau, line 18:
```suggestion
  netplayDolphinVersion: null as string | null,
  playbackDolphinVersion: null as string | null,
```
These should be defaulted to `null`. You're giving more responsibility to this value than it should have. It should store the value of the version if it exists, otherwise it should be null.

What you're trying to do with distinguishing between whether Dolphin has been installed or not is already provided by the DolphinStatus. If `DolphinStatus` is UNKNOWN then that means the complete handler never completed. That would give you the "not found" status you're trying to show. Additionally the string that is shown for the version (e.g. "Not Found") should be set in the renderer component not the store.

src/dolphin/install/installation.ts
vinceau: We shouldn't know or care where and when `installation.getDolphinVersion()` is or isn't called. If we're exposing a public function to get the dolphin version, I would expect it to return the version if it exists and null if it doesn't, rather than throwing. The caller shouldn't have to catch the error, null should indicate the installation doesn't exist.

src/dolphin/install/installation.ts
```diff
@@ -176,11 +176,16 @@ export class DolphinInstallation {
...
+    return !dolphinVersion || lt(dolphinVersion, latestVersion);
+  }
+
+  public async getDolphinVersion(): Promise<string | null> {
     const dolphinPath = await this.findDolphinExecutable();
     const dolphinVersionOut = spawnSync(dolphinPath, ["--version"]).stdout.toString();
     const match = dolphinVersionOut.match(semverRegex);
     const dolphinVersion = match?.[0] ?? "";
-    return lt(dolphinVersion, latestVersion);
+    return dolphinVersion || null;
```
vinceau, line 188:
```suggestion
    return dolphinVersion;
```
You should change the previous line to `match?.[0] ?? null` instead of changing the return.

src/renderer/containers/Settings/DolphinSettings.tsx
```diff
@@ -68,6 +71,11 @@ export const DolphinSettings: React.FC<{ dolphinType: DolphinLaunchType }> = ({
   return (
     <div>
       <Typography variant="h5">{dolphinTypeName} Dolphin Settings</Typography>
+      <Typography variant="caption">
+        Version: {!dolphinVersion && "Unknown"}
```
vinceau, line 75: If you wanted to distinguish between Dolphin not being found and the version being unknown, you should check the status of dolphin and display not found if the status is unknown.
e.g. something like: `const versionString = status === unknown ? "Not found" : !version ? "Unknown" : version;`

vinceau (changes requested):

vinceau (approved):

PR #354, opened 2022-12-11
[PR] Accept rules and policies

JLaferri:


src/renderer/lib/hooks/useQuickStart.ts
```diff
@@ -64,6 +72,7 @@ export const useQuickStart = () => {
     hasIso: Boolean(savedIsoPath),
     hasVerifiedEmail: Boolean(user?.emailVerified),
     hasPlayKey: Boolean(playKey),
+    rulesAccepted: playKey?.rulesAccepted ?? 0,
```
JLaferri, line 75: I realize the playKey is not being used as intended here. Though generally how I would expect things to work is that there would be an apollo store that would fetch all user info up-front. Not really sure how to achieve this at the moment.

NikhilNarayana: i don't think its a big deal. one thing I do wonder is if it would be better to do the boolean check here instead of in the if statement.

src/renderer/services/slippi/slippi.service.mock.ts
```diff
@@ -56,6 +57,11 @@ class MockSlippiBackendClient implements SlippiBackendService {
     await this.authService.updateDisplayName(name);
   }
 
+  @delayAndMaybeError(SHOULD_ERROR)
+  public async acceptRules() {
+    // Do nothing? Not sure how this should work
```
NikhilNarayana, line 62: just `// Do nothing` is fine since this is a server side action. just need to set the rulesAccepted value to 1 so it skips this step once we are done testing.

src/renderer/containers/QuickStart/VerifyEmailStep.tsx
```diff
@@ -104,7 +105,7 @@ export const VerifyEmailStep: React.FC = () => {
         Check Verification
       </Button>
       <div css={classes.emailNotFoundContainer}>
-        Not finding email?{" "}
+        Not finding email? Check spam folder. Still missing?{" "}
```
NikhilNarayana, line 108:
```suggestion
        Can't find the email? Check your spam folder. Still missing?{" "}
```
bit more verbose, but sounds better than what is currently there. could drop `verification` for brevity.

src/dolphin/playkey.ts
```diff
@@ -8,7 +8,17 @@ import type { DolphinInstallation } from "./install/installation";
 
 export async function writePlayKeyFile(installation: DolphinInstallation, playKey: PlayKey): Promise<void> {
   const keyPath = await findPlayKey(installation);
-  const contents = JSON.stringify(playKey, null, 2);
+  const contents = JSON.stringify(
```
NikhilNarayana, line 11: this is a bit unnecessary, just change the type signature to

`playKey: Omit<PlayKey, "rulesAccepted">`

NikhilNarayana (changes requested):

src/dolphin/playkey.ts
JLaferri: I didn't want to do this because if we add more fields, we'll have to remember to add them to this list.

src/common/constants.ts
```diff
@@ -6,6 +6,8 @@ export const slippiHomepage = "https://slippi.gg";
 export const slippiActivationUrl = "https://slippi.gg/online/enable";
 export const slippiManagePage = "https://slippi.gg/manage";
 
+export const currentRulesVersion = 1;
```
vinceau, line 9: Why does the current rules version need to be stored in the launcher? The backend knows what the latest version is, and knows the latest version the user has accepted. The only thing the frontend needs to know is whether to show the rules step (a boolean). Can't the backend just return a boolean to the frontend?

src/renderer/services/slippi/graphqlEndpoints.ts
```diff
@@ -38,7 +39,7 @@ export const QUERY_VALIDATE_USER_ID: TypedDocumentNode<
 
 export const QUERY_GET_USER_KEY: TypedDocumentNode<
```
vinceau, line 40: This should be renamed since we're now fetching more than just the key and potentially anything related to the user.

```suggestion
export const QUERY_GET_USER_DATA: TypedDocumentNode<
```

src/common/constants.ts
JLaferri: Hmmm maybe. Just a couple things at the moment though:

1. I don't want to think too hard about it right now.
2. The one issue I can think of is that let's say we update the slippi rules and want people to accept them again, if they haven't updated their launcher and the backend says they need to see the rules again, they would see and accept an old version of the rules

NikhilNarayana (approved):

PR #357, opened 2022-12-15
macOS: Mirror updated paths in Ishiiruka/374

ryanmcgrath:
This updates the Launcher to use the macOS `User` folder path(s) specified in [Ishiiruka/#374](https://github.com/project-slippi/Ishiiruka/pull/374).

The gist of this is that it appears that Ventura actually enforces signed bundles not being altered, and coupled with GateKeeper checking on each open as opposed to the first open, means that users get a message saying the app is damaged due to `Dolphin.ini`.

As far as I can tell this is the only core part that needs to be altered, and I went ahead and just mostly copied the approach the Linux block uses - but I defer to y'all.

### DIFF
--- a/src/dolphin/install/macos.ts
+++ b/src/dolphin/install/macos.ts
@@ -7,17 +7,19 @@ import { extractDmg } from "./extractDmg";
 export async function installDolphinOnMac({
   assetPath,
   destinationFolder,
+  skipUserFolderBkp,
   log = console.log,
 }: {
   assetPath: string;
   destinationFolder: string;
+  skipUserFolderBkp: boolean;
   log?: (message: string) => void;
 }) {
   const backupLocation = destinationFolder + "_old";
   const dolphinResourcesPath = path.join(destinationFolder, "Slippi Dolphin.app", "Contents", "Resources");
 
-  const alreadyInstalled = await fs.pathExists(dolphinResourcesPath);
-  if (alreadyInstalled) {
+  const bkpUserFolder = !skipUserFolderBkp && (await fs.pathExists(dolphinResourcesPath));
+  if (bkpUserFolder) {
     log(`${dolphinResourcesPath} already exists. Moving...`);
     await fs.move(destinationFolder, backupLocation, { overwrite: true });
   }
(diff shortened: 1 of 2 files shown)


vinceau (approved):
This change looks fine, but could this break existing users config/settings?

ryanmcgrath:
Yup! There's definitely a larger question of how to deploy this, lol.

NikhilNarayana:
@vinceau @ryanmcgrath i added a commit to handle the migration within the launcher itself cause doing it in dolphin was proving to be a pain. please re-review when you have some time.

src/dolphin/install/macos.ts
```diff
@@ -7,17 +7,19 @@ import { extractDmg } from "./extractDmg";
 export async function installDolphinOnMac({
   assetPath,
   destinationFolder,
+  skipUserFolderBkp,
   log = console.log,
 }: {
   assetPath: string;
   destinationFolder: string;
+  skipUserFolderBkp: boolean;
```
vinceau, line 15:
```suggestion
  shouldBackupUserFolder?: boolean;
```
This is an exposed type so we shouldn't be using short-names. Can we also change the prop so that we have it true when we want to backup instead of having it true to skip back up and then you can just set the default appropriately.

vinceau (changes requested):

src/dolphin/install/macos.ts
```diff
@@ -7,17 +7,19 @@ import { extractDmg } from "./extractDmg";
...
+  shouldBackupUserFolder: boolean;
   log?: (message: string) => void;
 }) {
   const backupLocation = destinationFolder + "_old";
   const dolphinResourcesPath = path.join(destinationFolder, "Slippi Dolphin.app", "Contents", "Resources");
 
-  const alreadyInstalled = await fs.pathExists(dolphinResourcesPath);
-  if (alreadyInstalled) {
+  const bkpUserFolder = shouldBackupUserFolder && (await fs.pathExists(dolphinResourcesPath));
+  if (bkpUserFolder) {
```
vinceau, line 22:
```suggestion
  const backupUserFolder = shouldBackupUserFolder && (await fs.pathExists(dolphinResourcesPath));
  if (backupUserFolder) {
```
nit: can we just call this backup for consistency?

vinceau (approved):
LGTM with these minor nits

PR #358, opened 2022-12-15
feat: add error when email not yet verified

JLaferri:
before I think people were clicking the button without reading and having no idea whats going on when the step wasnt progressing

image

### DIFF
--- a/src/renderer/containers/QuickStart/VerifyEmailStep.tsx
+++ b/src/renderer/containers/QuickStart/VerifyEmailStep.tsx
@@ -82,7 +82,7 @@ export const VerifyEmailStep: React.FC = () => {
       // Get current user manually since the user variable above hasn't updated yet
       const newUser = authService.getCurrentUser();
       if (!newUser?.emailVerified) {
-        showError("Email still not verified. Check spam folder?");
+        showError("Email is not yet verified. Have you checked your spam folder?");
       }
     } catch (err: any) {
       showError(err.message);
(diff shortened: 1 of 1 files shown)


src/renderer/containers/QuickStart/VerifyEmailStep.tsx
```diff
@@ -76,9 +76,17 @@ export const VerifyEmailStep: React.FC = () => {
...
 
   const handleCheckVerification = async () => {
-    authService.refreshUser().catch((err) => {
+    try {
+      await authService.refreshUser();
+
+      // Get current user manually since the user variable above hasn't updated yet
+      const newUser = authService.getCurrentUser();
+      if (!newUser?.emailVerified) {
+        showError("Email still not verified. Check spam folder?");
```
vinceau, line 85:
```suggestion
        showError("Email is not yet verified. Have you checked your spam folder?");
```
I think I'd prefer a more fuller sentence here.

vinceau (approved):
LGTM but like yeah I guess I'd prefer a fuller error message.

src/renderer/containers/QuickStart/VerifyEmailStep.tsx
JLaferri: I made it shorter to prevent the multi-line, hoping people would actually read it lol.

NikhilNarayana:
```suggestion
        showError("Email still not verified. Checked spam folder?");
```
my only suggestion is to imply "[have you] checked [your] spam folder?"

or go with a more direct approach
```suggestion
        showError("Email still not verified. Check your spam folder");
```

vinceau: I just don't think adding an extra 5 words or so is going to be the deciding factor of whether people read the prompt or not.

JLaferri: It doesn’t matter much to me. I think it’s far more important we have some kind of error and get it live quickly rather than deliberating on it.

Just change it to whatever and push it imo.

PR #361, opened 2022-12-24
feat: add button to open dolphin settings folder 

NikhilNarayana:
Swap the `Open containing folder` with `Open settings folder` which opens the correct location for where the User settings are stored on each OS. for windows this is just where dolphin is located, but on macOS/linux it is a separate config folder.
image

### DIFF
--- a/src/dolphin/setup.ts
+++ b/src/dolphin/setup.ts
@@ -1,4 +1,4 @@
-import { app } from "electron";
+import { app, shell } from "electron";
 import log from "electron-log";
 import * as fs from "fs-extra";
 import { isEqual } from "lodash";
@@ -11,10 +11,10 @@ import {
   ipc_configureDolphin,
   ipc_dolphinEvent,
   ipc_downloadDolphin,
-  ipc_getUserFolderPath,
   ipc_hardResetDolphin,
   ipc_importDolphinSettings,
   ipc_launchNetplayDolphin,
+  ipc_openDolphinSettingsFolder,
   ipc_removePlayKeyFile,
   ipc_softResetDolphin,
   ipc_storePlayKeyFile,
@@ -50,9 +50,10 @@ export default function setupDolphinIpc({ dolphinManager }: { dolphinManager: Do
     return { success: true };
   });
 
-  ipc_getUserFolderPath.main!.handle(async ({ dolphinType }) => {
+  ipc_openDolphinSettingsFolder.main!.handle(async ({ dolphinType }) => {
     const path = await dolphinManager.getInstallation(dolphinType).userFolder;
-    return { path };
+    await shell.openPath(path);
+    return { success: true };
   });
 
   ipc_hardResetDolphin.main!.handle(async ({ dolphinType }) => {
--- a/src/dolphin/setup.ts
+++ b/src/dolphin/setup.ts
@@ -11,6 +11,7 @@ import {
   ipc_configureDolphin,
   ipc_dolphinEvent,
   ipc_downloadDolphin,
+  ipc_getUserFolderPath,
   ipc_hardResetDolphin,
   ipc_importDolphinSettings,
   ipc_launchNetplayDolphin,
@@ -47,6 +48,11 @@ export default function setupDolphinIpc({ dolphinManager }: { dolphinManager: Do
     console.log("soft resetting dolphin...");
     await dolphinManager.reinstallDolphin(dolphinType, false);
     return { success: true };
+  });
+
+  ipc_getUserFolderPath.main!.handle(async ({ dolphinType }) => {
+    const path = await dolphinManager.getInstallation(dolphinType).userFolder;
+    return { path };
   });
 
   ipc_hardResetDolphin.main!.handle(async ({ dolphinType }) => {
--- a/src/main/menu.ts
+++ b/src/main/menu.ts
@@ -76,7 +76,7 @@ export class MenuBuilder {
           },
         },
         {
-          label: "Open Application Support Folder",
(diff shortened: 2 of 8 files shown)


vinceau (commented):
It seems a bit cluttered with 3 buttons. Does the `Open containing folder` button actually provide much value on Mac and Linux? Maybe @ryanmcgrath can chime in for the Mac side.

NikhilNarayana:
I see no real value on Linux.

On macOS I see some value in being able to check if there are any bundle issues, but otherwise I don't see a need. I would also like Ryan's input.

vinceau:
If it's primarily only for debugging purposes the path to the app bundle is also shown directly underneath in the text field. So for the average user they might not need.

ryanmcgrath:
Debugging is probably the big case, yeah. You could add it as a menubar sub item if you didn't want to populate a button in the app?

NikhilNarayana:
possibly dumb idea, consolidate everything into one button and open both folders on macOS/Linux.
image

src/renderer/lib/dolphin/useDolphinActions.ts
```diff
@@ -23,6 +23,13 @@ export const useDolphinActions = (dolphinService: DolphinService) => {
     [netplayStatus, playbackStatus],
   );
 
+  const getDolphinUserPath = useCallback(
+    async (dolphinType: DolphinLaunchType): Promise<string> => {
+      return await dolphinService.getUserFolderPath(dolphinType);
+    },
+    [dolphinService],
+  );
+
```
vinceau, line 32: Hmm I don't think we don't really need this function. The caller could just call `dolphinService.getUserFolderPath()` directly, right?

src/renderer/containers/Settings/DolphinSettings.tsx
```diff
@@ -38,13 +38,18 @@ export const DolphinSettings: React.FC<{ dolphinType: DolphinLaunchType }> = ({
...
   const dolphinIsReady = dolphinStatus === DolphinStatus.READY && !dolphinIsOpen && isResetType === null;
   const versionString: string =
     dolphinStatus === DolphinStatus.UNKNOWN ? "Not found" : !dolphinVersion ? "Unknown" : dolphinVersion;
 
   const openDolphinDirectoryHandler = async () => {
+    if (isMac || isLinux) {
+      const userSettingsPath = await getDolphinUserPath(dolphinType);
+      await window.electron.shell.openPath(userSettingsPath);
+    }
     await window.electron.shell.openPath(dolphinPath);
```
vinceau, line 53: Wait so on mac and linux it will open two folders? I don't really like that. Can we just open one folder, and call it `Open settings folder`?
```suggestion
      await window.electron.shell.openPath(userSettingsPath);
    } else {
      await window.electron.shell.openPath(dolphinPath);
    }
```

src/dolphin/setup.ts
```diff
@@ -49,6 +50,11 @@ export default function setupDolphinIpc({ dolphinManager }: { dolphinManager: Do
     return { success: true };
   });
 
+  ipc_getUserFolderPath.main!.handle(async ({ dolphinType }) => {
```
vinceau, line 53:
```suggestion
  ipc_openDolphinSettingsFolder.main!.handle(async ({ dolphinType }) => {
```
What are your thoughts on making this IPC call in charge of opening the folder too? And then it can switch on the OS and open the appropriate folder. I feel like it's like a extra hoop to jump to request the folder via IPC, and then once again call the IPC to open the folder. Can you foresee any other uses for the FE to get the user folder?

vinceau (changes requested):
I think we should think about this a little bit more. Can you also update the PR description? I think things have changed a bit since then.

src/dolphin/setup.ts
NikhilNarayana: yeah that makes more sense

src/renderer/containers/Settings/DolphinSettings.tsx
NikhilNarayana: i still worry about not being able to easily access the bundle folder :/ i really want to avoid making support difficult. I would honestly take anything that keeps it available. Maybe even a context menu on the button haha (this is a joke don't take this seriously)

src/main/menu.ts
```diff
@@ -72,6 +75,12 @@ export class MenuBuilder {
             this.onOpenPreferences();
           },
         },
+        {
+          label: "Open Application Support Folder",
```
vinceau, line 79: This is looking pretty good but I think it's still not very clear what this should do for the end user. Is this the `Application Support` folder for the launcher? Is it for the `Application Support` folder for Dolphin? If it's for Dolphin, which Dolphin? Netplay or Playback? I just don't think it's very clear what this should do and when it should be used.

ryanmcgrath: You might consider just "Open Slippi Dolphin Application Support Folder" or something, since Slippi Dolphin shares a common folder with subdirectories for netplay and playback-specific stuff.

ryanmcgrath: Ah, question then: am I misunderstanding what you want to solve here?

e.g, user currently digs into Slippi Dolphin bundle to do x/y/z, that would now need to dig into `~/Library/Application\ Support/com.project-slippi.dolphin/` - this, as it stands, would be opening `~/Library/Application\ Support/Slippi\ Launcher/...`, no? This is assuming this change is combined with having Dolphin extract the folder on-first-run.

(Or am I conflating multiple unrelated things here?)

vinceau (approved):

PR #362, opened 2022-12-24
feat: add check for updates button

NikhilNarayana:
The slippi logo in the top right is repurposed to be a check for updates button. failures during an update aren't handled atm.

image

### DIFF
--- a/src/renderer/components/Footer.tsx
+++ b/src/renderer/components/Footer.tsx
@@ -10,6 +10,7 @@ import React from "react";
 import { ExternalLink as A } from "@/components/ExternalLink";
 import { useDolphinActions } from "@/lib/dolphin/useDolphinActions";
 import { useAppUpdate } from "@/lib/hooks/useAppUpdate";
+import { useToasts } from "@/lib/hooks/useToasts";
 import { useServices } from "@/services";
 import { ReactComponent as DiscordIcon } from "@/styles/images/discord.svg";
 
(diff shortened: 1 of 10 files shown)


src/renderer/components/PersistentNotification.tsx
```diff
@@ -5,13 +5,29 @@ import ButtonBase from "@mui/material/ButtonBase";
...
+  if (!updateReady && updateDownloadProgress) {
+    return (
+      <Outer>
+        <div
+          css={css`
+            display: flex;
+            justify-content: center;
+          `}
+        >
+          Downloading Version {updateVersion}...
```
vinceau, line 26:
```suggestion
          Downloading version {updateVersion}...
```
nit: i think i'd prefer sentence case here but i'm also not super fussed.

src/renderer/lib/hooks/useAppUpdate.ts
```diff
@@ -1,8 +1,16 @@
 import { useToasts } from "@/lib/hooks/useToasts";
 
-export const useInstallAppUpdate = () => {
+export const useAppUpdate = () => {
   const { showError } = useToasts();
 
+  const checkForAppUpdates = async () => {
```
vinceau, line 6:
```suggestion
  const checkForAppUpdates = React.useCallback(async () => {
```
Do you mind wrapping both of these in `useCallback` for me?

src/renderer/components/Footer.tsx
```diff
@@ -26,8 +31,40 @@ export const BasicFooter = styled.div`
...
+    } catch (err) {
+      window.electron.log.error(err);
+      showError("Failed to get updates");
+    }
+    setTimeout(() => setCheckingForUpdates(false), 10000);
+  }, [checkForAppUpdates, updateDolphin, showInfo, showError]);
+
   return (
     <Outer>
+      <Button
```
vinceau, line 54: Instead of introducing yet another button, let's repurpose the Slippi logo in the top right for checking for updates. I doubt anyone has ever clicked that intentionally.

vinceau (changes requested):

src/renderer/components/Footer.tsx
```diff
@@ -26,8 +31,40 @@ export const BasicFooter = styled.div`
...
+    setCheckingForUpdates(true);
+    try {
+      showInfo("Checking for updates...");
+      await checkForAppUpdates();
+      await updateDolphin();
+    } catch (err) {
+      window.electron.log.error(err);
+      showError("Failed to get updates");
+    }
+    setTimeout(() => setCheckingForUpdates(false), 10000);
```
vinceau, line 49:
```suggestion
    } finally {
      setCheckingForUpdates(false);
    }
```
Not sure what the `setTimeout` here is for, should this be a `finally`?

NikhilNarayana: the reason i put that there is to avoid people spamming the button cause
```
await checkForAppUpdates();
await updateDolphin();
```
return almost immediately rather than after the checks are done.

vinceau: That's almost certainly not the correct solution to the problem. To only trigger the event once, upon button spam completion you can use debounce. To trigger the event at most once every X during button spam you can use throttle. I think it's still probably correct to set the `setCheckingForUpdates` back to false on `finally` though.

vinceau (approved):

PR #363, opened 2022-12-25
feat: show winners in replay browser

NikhilNarayana:
just enabling what vince already setup
image

### DIFF
--- a/src/replays/loadFile.ts
+++ b/src/replays/loadFile.ts
@@ -26,10 +26,7 @@ export async function loadFile(fullPath: string): Promise<FileResult> {
     winners: [],
   };
 
-  const winners = game.getWinners();
-  if (winners) {
-    result.winners = winners.map((winner) => winner.playerIndex);
-  }
+  result.winners = game.getWinners().map((winner) => winner.playerIndex);
 
   // Load metadata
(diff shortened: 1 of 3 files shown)


src/replays/loadFile.ts
```diff
@@ -23,8 +23,14 @@ export async function loadFile(fullPath: string): Promise<FileResult> {
     startTime: null,
     lastFrame: null,
     metadata: null,
+    winners: null,
   };
 
+  const winners = game.getWinners();
```
vinceau, line 29:
```suggestion
  const winners = game.getWinners();
```
this should always exist or at least be an empty array if unsupported so you don't need the following check.

vinceau (changes requested):

src/replays/types.ts
```diff
@@ -7,6 +7,7 @@ export interface FileResult {
   startTime: string | null;
   lastFrame: number | null;
   metadata: MetadataType | null;
+  winners: number[];
```
vinceau, line 10:
```suggestion
  winnerIndices: number[];
```
last nit: can we call these indices so it's a bit clearer as to what the number is. like it's not port but player index.

vinceau (approved):
LGTM once the last nit is addresed

PR #365, opened 2022-12-26
fix: extend SLP directory path validation to be more robust

### LINKED ISSUE #359: Extra SLP directories with substring of root directory can't be added
Attempting to add an adjacent directory that contains the root directory path a substring will result in `Cannot add sub directories of the Root SLP Directory.` error. Users should be able to add these directories since there isn't a real conflict.

Example:
Root = `C:\Users\Nikki\Documents\Slippi`
Adding `C:\Users\Nikki\Documents\Slippi Test` will result in the error mentioned above.

dryu99:
Resolves #359

### Description

Previously, attempting to add an additional SLP directory that:
1. was adjacent to the root SLP directory
2. in its name, contained a substring of the root SLP directory name

...would result in a `Cannot add sub directories of the Root SLP Directory` error. 

e.g.
Root = `C:\Users\Nikki\Documents\Slippi`
Adding `C:\Users\Nikki\Documents\Slippi Test` causes an error.

This was occurring because the previous SLP directory path validation used [`includes`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/includes) to check whether the new path was a subdirectory of the root path.

This PR fixes the issue by leveraging an existing utils function that checks whether a folder is a subdirectory of another.

Screen Shot 2022-12-26 at 6 52 22 AM

### Testing

- Create a directory adjacent to the root SLP directory and name it such that it contains a substring of the root SLP directory name (e.g. "Slippi Test" if your root directory is "Slippi")
- Add the directory as an additional SLP directory via Slippi
  - [ ] Verify that no error is displayed and the directory is added successfully
- After adding the directory, create a child directory inside of it and try adding the child as an additional SLP directory
  - [ ] Verify that an error is displayed and the directory is not added
- Remove the parent directory via Slippi
- Add the child directory as an additional SLP directory via Slippi
  - [ ] Verify that no error is displayed and the directory is added successfully
- Add the parent directory as an additional SLP directory via Slippi
  - [ ] Verify that the displayed path changes from the child path to the parent path

### DIFF
--- a/src/main/preload.ts
+++ b/src/main/preload.ts
@@ -6,7 +6,7 @@ import settingsApi from "@settings/api";
 import { clipboard, contextBridge, shell } from "electron";
 import log from "electron-log";
 import path from "path";
-import isSubdirectory from "utils/isSubdirectory";
+import { isSubdirectory } from "utils/isSubdirectory";
 
 import commonApi from "./api";
 
--- a/src/main/preload.ts
+++ b/src/main/preload.ts
@@ -6,6 +6,7 @@ import settingsApi from "@settings/api";
 import { clipboard, contextBridge, shell } from "electron";
 import log from "electron-log";
 import path from "path";
+import isSubdirectory from "utils/isSubdirectory";
 
 import commonApi from "./api";
 
@@ -16,10 +17,11 @@ const api = {
   broadcast: broadcastApi,
   dolphin: dolphinApi,
   replays: replaysApi,
+  utils: {
+    isSubdirectory,
+  },
   path: {
     join: path.join,
-    isAbsolute: path.isAbsolute,
-    relative: path.relative,
   },
   clipboard: {
     writeText: clipboard.writeText,
--- a/src/main/preload.ts
+++ b/src/main/preload.ts
@@ -18,6 +18,8 @@ const api = {
   replays: replaysApi,
   path: {
     join: path.join,
+    isAbsolute: path.isAbsolute,
+    relative: path.relative,
   },
   clipboard: {
     writeText: clipboard.writeText,
--- a/src/renderer/components/MultiPathInput.tsx
+++ b/src/renderer/components/MultiPathInput.tsx
@@ -8,7 +8,7 @@ import React, { useState } from "react";
 import { useSettings } from "@/lib/hooks/useSettings";
 import { useToasts } from "@/lib/hooks/useToasts";
 
-const utils = window.electron.utils;
+const { isSubdirectory } = window.electron.utils;
 
 export interface MultiPathInputProps {
   updatePaths: (paths: string[]) => void;
--- a/src/renderer/components/MultiPathInput.tsx
+++ b/src/renderer/components/MultiPathInput.tsx
@@ -7,7 +7,8 @@ import React, { useState } from "react";
 
 import { useSettings } from "@/lib/hooks/useSettings";
 import { useToasts } from "@/lib/hooks/useToasts";
-import { isSubdirectory } from "@/lib/utils";
+
+const utils = window.electron.utils;
 
 export interface MultiPathInputProps {
   updatePaths: (paths: string[]) => void;
@@ -28,7 +29,7 @@ export const MultiPathInput: React.FC<MultiPathInputProps> = ({ paths, updatePat
       return false;
     }
 
-    if (isSubdirectory(rootFolder, newPath)) {
+    if (utils.isSubdirectory(rootFolder, newPath)) {
       addErrorToast("Cannot add sub directories of the Root SLP Directory.");
       return false;
     }
@@ -36,10 +37,10 @@ export const MultiPathInput: React.FC<MultiPathInputProps> = ({ paths, updatePat
     let pathsToCheck = paths;
     for (let i = 0; i < pathsToCheck.length; i++) {
       const path = pathsToCheck[i];
-      if (isSubdirectory(path, newPath)) {
+      if (utils.isSubdirectory(path, newPath)) {
         addErrorToast("Cannot add sub directories of the Root SLP Directory.");
         return false;
-      } else if (isSubdirectory(newPath, path)) {
+      } else if (utils.isSubdirectory(newPath, path)) {
         updatePaths(pathsToCheck.splice(i, 1));
         pathsToCheck = pathsToCheck.splice(i--, 1); //decrement i because we are dropping an entry
       }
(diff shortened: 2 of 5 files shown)


src/renderer/components/MultiPathInput.tsx
```diff
@@ -27,18 +28,18 @@ export const MultiPathInput: React.FC<MultiPathInputProps> = ({ paths, updatePat
...
+    if (isSubdir(rootFolder, newPath)) {
       addErrorToast("Cannot add sub directories of the Root SLP Directory.");
       return false;
     }
 
     let pathsToCheck = paths;
     for (let i = 0; i < pathsToCheck.length; i++) {
       const path = pathsToCheck[i];
-      if (newPath.includes(path)) {
+      if (isSubdir(path, newPath)) {
```
dryu99, line 39: Opted to also call `isSubdir` in other spots where the subdirectory check was being done. The tests should validate that the functionality still works as it originally did.

src/renderer/lib/utils.ts
```diff
@@ -61,3 +62,12 @@ export const humanReadableBytes = (bytes: number): string => {
...
 };
+
+export const isSubdir = (parentPath: string, dirPath: string) => {
+  const relativePath = path.relative(parentPath, dirPath);
+  return (
+    relativePath && // value is "" when paths are identical
+    !relativePath.startsWith("..") &&
+    !path.isAbsolute(relativePath) // covers case where different drives are used e.g. 'C:\\A' vs 'D:\\A\\B'
+  );
+};
```
dryu99, line 73: This logic isn't perfect, as there's still the [edge case where a user creates a directory that starts with `".."`](https://stackoverflow.com/a/45242825). However because this seems unlikely, I think it would be okay to settle with this for now. Open to hearing other people's thoughts on this though!

src/renderer/lib/utils.ts
```diff
@@ -61,3 +62,12 @@ export const humanReadableBytes = (bytes: number): string => {
 
   return `0 ${sizes[0]}`;
 };
+
+export const isSubdir = (parentPath: string, dirPath: string) => {
+  const relativePath = path.relative(parentPath, dirPath);
+  return (
+    relativePath && // value is "" when paths are identical
```
NikhilNarayana, line 69: lets avoid throwing comments in line like this. rather put them together in one block before the return or create variables and add the comments there.

src/renderer/lib/utils.ts
NikhilNarayana: I think that's fine, not the worst situation to deal with.

NikhilNarayana (changes requested):
one minor nit, otherwise i think it looks good

src/renderer/lib/utils.ts
```diff
@@ -1,4 +1,5 @@
 import { characters as charUtils, stages as stageUtils } from "@slippi/slippi-js";
+import path from "path";
```
vinceau, line 2: `path` is a node built in which you're importing from the web renderer. you should instead expose required node modules via the context bridge.

src/renderer/lib/utils.ts
```diff
@@ -61,3 +62,12 @@ export const humanReadableBytes = (bytes: number): string => {
 
   return `0 ${sizes[0]}`;
 };
+
+export const isSubdir = (parentPath: string, dirPath: string) => {
```
vinceau, line 66: can we rename this to `isSubdirectory` to avoid short names in public exported functions and interfaces?

vinceau (changes requested):

src/renderer/lib/utils.ts
dryu99: Didn't know about context bridges, thanks for pointing that out.

dryu99:
Sorry for the request review spam, I was on a plane and the wifi was being really spotty 😅

vinceau (changes requested):
So I just remembered that we already have an `isSubdirectory` function in `src/replays/folderTreeService.ts`. Here's what I think we should do to avoid duplication:

1. take the existing function in the codebase and move it to `src/utils/isSubdirectory.ts`
2. update `folderTreeService.ts` to import from utils
3. expose the `isSubdirectory` function over context bridge instead of all those path helper functions

src/utils/isSubdirectory.ts
```diff
@@ -0,0 +1,10 @@
+import path from "path";
+
+// Taken from: https://stackoverflow.com/questions/37521893/determine-if-a-path-is-subdirectory-of-another-in-node-js
+const isSubdirectory = (parent: string, dir: string): boolean => {
```
vinceau, line 4:
```suggestion
export const isSubdirectory = (parent: string, dir: string): boolean => {
```

src/utils/isSubdirectory.ts
```diff
@@ -0,0 +1,10 @@
...
+  const relative = path.relative(parent, dir);
+  const isSubdir = Boolean(relative) && !relative.startsWith("..") && !path.isAbsolute(relative);
+  return isSubdir;
+};
+
+export default isSubdirectory;
```
vinceau, line 10:
```suggestion
```

src/replays/folderTreeService.ts
```diff
@@ -1,5 +1,6 @@
 import * as fs from "fs-extra";
 import path from "path";
+import isSubdirectory from "utils/isSubdirectory";
```
vinceau, line 3:
```suggestion
import { isSubdirectory } from "utils/isSubdirectory";
```

src/renderer/components/MultiPathInput.tsx
```diff
@@ -8,6 +8,8 @@ import React, { useState } from "react";
 import { useSettings } from "@/lib/hooks/useSettings";
 import { useToasts } from "@/lib/hooks/useToasts";
 
+const utils = window.electron.utils;
```
vinceau, line 11:
```suggestion
const { isSubdirectory } = window.electron.utils;
```

src/main/preload.ts
```diff
@@ -6,6 +6,7 @@ import settingsApi from "@settings/api";
 import { clipboard, contextBridge, shell } from "electron";
 import log from "electron-log";
 import path from "path";
+import isSubdirectory from "utils/isSubdirectory";
```
vinceau, line 9:
```suggestion
import { isSubdirectory } from "utils/isSubdirectory";
```

vinceau (changes requested):
Looking really good, just a few more nits. We prefer to avoid `export default` unless there's a specific reason why we're using it.

vinceau (approved):
LGTM! Thanks for taking this on. :)

NikhilNarayana (approved):

PR #367, opened 2022-12-30
Prevent Electron window title changes resulting from navigation.

### LINKED ISSUE #366: Using mouse back/forward buttons changes window title to 'index.html'
Steps to repro:
1) Click any of the top tabs (e.g. Replays)
2) Press back button on mouse

Expected: title of the window remains "Slippi Launcher"
Actual: title becomes "index.html"

tec27:
Fixes #366.

This is one of the ways to fix this, there are others that may be preferable (such as doing `preventDefault` on title changes in the main process). If you'd like a different approach (or to pull this title from the product name during the webpack build or something), let me know, happy to change it!

### DIFF
--- a/src/main/main.ts
+++ b/src/main/main.ts
@@ -118,6 +118,11 @@ const createWindow = async () => {
     mainWindow = null;
   });
 
+  mainWindow.on("page-title-updated", (event) => {
+    // Always keep the initial window title
+    event.preventDefault();
+  });
+
   const menuBuilder = new MenuBuilder({
     mainWindow,
     onOpenPreferences: () => {


vinceau:
Can you provide an example of how you might pull this from the product name rather than hard coding it in the EJS?

tec27:
So to make that happen, the title tag in `index.ejs` would become:

```
<title><%= title %></title>
```

and both webpack configurations (prod and dev) would update their `HtmlWebpackPlugin` call to include that parameter:
```ts
// ...
    new HtmlWebpackPlugin({
      filename: "index.html",
      template: path.join(webpackPaths.srcRendererPath, "index.ejs"),
      templateParameters: {
        title: productName,
      },
      minify: {
        collapseWhitespace: true,
        removeAttributeQuotes: true,
        removeComments: true,
      },
      isBrowser: false,
      isDevelopment: process.env.NODE_ENV !== "production",
    }),
// ...
```

Would need to read `productName` out of `release/app/package.json` though which may be slightly annoying (either need to `require` it or do `fs.readFile` + `JSON.parse`).

---------------

If the goal is to just keep the initial title always (which seems to be pulled from `productName` already currently), the easier thing might be to just add an event listener on [mainWindow](https://github.com/project-slippi/slippi-launcher/blob/8e8d458d75d2a5374f69012b65a59129055d49c4/src/main/main.ts#L84):

```ts
mainWindow.addEventListener('page-title-updated', event => {
  // Always keep the initial window title
  event.preventDefault();
});
```

vinceau:
> So to make that happen, the title tag in `index.ejs` would become:
> 
> ```
> <title><%= title %></title>
> ```
> 
> and both webpack configurations (prod and dev) would update their `HtmlWebpackPlugin` call to include that parameter:
> 
> ```ts
> // ...
>     new HtmlWebpackPlugin({
>       filename: "index.html",
>       template: path.join(webpackPaths.srcRendererPath, "index.ejs"),
>       templateParameters: {
>         title: productName,
>       },
>       minify: {
>         collapseWhitespace: true,
>         removeAttributeQuotes: true,
>         removeComments: true,
>       },
>       isBrowser: false,
>       isDevelopment: process.env.NODE_ENV !== "production",
>     }),
> // ...
> ```
> 
> Would need to read `productName` out of `release/app/package.json` though which may be slightly annoying (either need to `require` it or do `fs.readFile` + `JSON.parse`).
> 
> If the goal is to just keep the initial title always (which seems to be pulled from `productName` already currently), the easier thing might be to just add an event listener on [mainWindow](https://github.com/project-slippi/slippi-launcher/blob/8e8d458d75d2a5374f69012b65a59129055d49c4/src/main/main.ts#L84):
> 
> ```ts
> mainWindow.on('page-title-updated', event => {
>   // Always keep the initial window title
>   event.preventDefault();
> });
> ```

Thanks for the example code. I think I'd prefer it to just `preventDefault` on the title change event.

tec27:
Cool, pushed a new commit with that change instead!

vinceau (approved):
LGTM. Thanks for looking into this.

PR #370, opened 2023-01-15
Pressing play twice while Slippi is in "Launch Dolphin" mode launches Melee directly

sticks-stuff:
A simple QoL change that makes it easier for people to get directly into Melee if they so choose

### DIFF
--- a/src/dolphin/manager.ts
+++ b/src/dolphin/manager.ts
@@ -79,17 +79,22 @@ export class DolphinManager {
   }
 
   public async launchNetplayDolphin() {
-    if (this.netplayDolphinInstance) {
-      throw new Error("Netplay dolphin is already open!");
-    }
-
     await this._updateDolphinSettings(DolphinLaunchType.NETPLAY);
 
     const netplayInstallation = this.getInstallation(DolphinLaunchType.NETPLAY);
     const dolphinPath = await netplayInstallation.findDolphinExecutable();
     log.info(`Launching dolphin at path: ${dolphinPath}`);
     const launchMeleeOnPlay = this.settingsManager.get().settings.launchMeleeOnPlay;
-    const meleeIsoPath = launchMeleeOnPlay ? await this._getIsoPath() : undefined;
+    let meleeIsoPath = launchMeleeOnPlay ? await this._getIsoPath() : undefined;
+
+    if (this.netplayDolphinInstance) {
+      if (!launchMeleeOnPlay) {
+        this.netplayDolphinInstance.kill();
+        meleeIsoPath = await this._getIsoPath();
+      } else {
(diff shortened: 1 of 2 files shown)


src/dolphin/manager.ts
```diff
@@ -79,17 +79,22 @@ export class DolphinManager {
...
-    if (this.netplayDolphinInstance) {
-      throw new Error("Netplay dolphin is already open!");
-    }
-
     await this._updateDolphinSettings(DolphinLaunchType.NETPLAY);
 
     const netplayInstallation = this.getInstallation(DolphinLaunchType.NETPLAY);
     const dolphinPath = await netplayInstallation.findDolphinExecutable();
     log.info(`Launching dolphin at path: ${dolphinPath}`);
     const launchMeleeOnPlay = this.settingsManager.get().settings.launchMeleeOnPlay;
```
vinceau, line 87:
```suggestion
    const launchMeleeOnPlay = this.settingsManager.get().settings.launchMeleeOnPlay || 
this.netplayDolphinInstance !== null;
```
I think this achieves what you want without needing to change `meleeIsoPath` later on yeah?

src/dolphin/manager.ts
```diff
@@ -79,17 +79,22 @@ export class DolphinManager {
...
-    const meleeIsoPath = launchMeleeOnPlay ? await this._getIsoPath() : undefined;
+    let meleeIsoPath = launchMeleeOnPlay ? await this._getIsoPath() : undefined;
+
+    if (this.netplayDolphinInstance) {
+      if (!launchMeleeOnPlay) {
+        this.netplayDolphinInstance.kill();
```
vinceau, line 92:
```suggestion
        this.netplayDolphinInstance.kill();
```
This sounds good in theory, but I worry that people could kill their dolphin process A) unintentionally (e.g. they're already in a game or in a training mode or something and we kill it), or B) intentionally by skirting around the game reporting logic in Dolphin (this is already an existing issue).

vinceau (commented):
Thanks for the PR. This is a good idea in theory but in practice there are more things that need ironing out.
1. We need to make sure Dolphin can still submit the game report when killed by the launcher
2. Ideally we can just launch Melee directly in the existing Dolphin instance, without quitting Dolphin and re-opening it again.

CC: @NikhilNarayana

sticks-stuff:
> Ideally we can just launch Melee directly in the existing Dolphin instance, without quitting Dolphin and re-opening it again.

This would be ideal however it seems beyond my skill level to do memory management stuff like that. External Melee Camera does have a way of calling native Dolphin functions (https://github.com/sadkellz/External-Melee-Camera/blob/main/External-Melee-Camera/emc_functions.py#L134) so maybe thats something to look into?

PR #375, opened 2023-03-08
feat: Make player tags in the replay browser clickable

trstruth:
Hello! I implemented a tiny feature whereby you can click a players name on the replay browser and it will open their profile on slippi.gg in your browser:

out

I picked this up after I saw it was a [feature request in the discord channel](https://discord.com/channels/328261477372919811/1074619651838971924)

This is my first time contributing to this project, so I did my best to follow examples of other PRs I saw in the repo but please let me know if I'm not properly following any conventions here

### DIFF
--- a/src/renderer/containers/ReplayBrowser/replay_file/team_elements/player_badge/PlayerBadge.tsx
+++ b/src/renderer/containers/ReplayBrowser/replay_file/team_elements/player_badge/PlayerBadge.tsx
@@ -1,9 +1,11 @@
 import { css } from "@emotion/react";
-import React from "react";
+import React, { useState } from "react";
 
 import { getColor } from "@/lib/playerColors";
 import { getCharacterIcon } from "@/lib/utils";
 import crownImage from "@/styles/images/crown.png";
+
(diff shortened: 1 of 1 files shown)


vinceau:
Hey, thanks for this PR! I think this is something we'd like to add in the future but this current implementation is not it. Here are some reasons why:
1. The inconsistent behaviour of some badges being clickable and some not is quite jarring.
2. It occupies the badge click functionality in case we wanted to use that behaviour for something else in the future.
3. The `PlayerBadge` which should be a dumb component now also has to handle the click function which is undesirable. I'd rather leave that to the parent.

You're welcome to try address some of these issues if you find yourself with time but both me and Nikki are very busy and may find it hard to give you sufficient feedback to get the PR in a state that is mergable.

Here are some suggestions for how I would see this feature implemented:
* Leave the player badge as a dumb component, and handle the on click in the parent
* On click, show a menu (like the overflow button) which gives you an option to `Show player profile` if it's a valid connect code. Maybe this option can be grayed out/disabled if it's not a connect code. This fixes 1 and 2 and since it's a menu, it also allows us to show more options in the future.

PR #377, opened 2023-03-17
Add Gecko Code handling

cnkeats:
This PR aims to implement various functions of interacting with Gecko Codes to cover #177
Obsoletes #201 which is out of date

<hr>

Feature List
- [x] Toggle existing codes on and off
- [x] Delete existing codes
  - Default codes can be disabled but not deleted
- [x] Add new codes
- [x] Copy codes to clipboard

<hr>

Polishing TODO
- [x] Toast on error
- [x] Inform user when code is copied
- [x] Prevent adding duplicate codes
- [x] UI improvements
  - Suggestions very welcome

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -62,7 +62,7 @@ module.exports = {
     "react/display-name": "off",
     "react/no-unescaped-entities": "warn",
     "@typescript-eslint/no-unused-vars": [
-      "error",
+      "off",
       {
         argsIgnorePattern: "^_",
         varsIgnorePattern: "^_",
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -62,7 +62,7 @@ module.exports = {
     "react/display-name": "off",
     "react/no-unescaped-entities": "warn",
     "@typescript-eslint/no-unused-vars": [
-      "off",
+      "error",
       {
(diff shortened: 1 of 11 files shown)


NikhilNarayana (changes requested):

vinceau (changes requested):
Thanks for taking this on! Just a few more comments. I think this title is too long.

image

and I'm also not a fan of this jump in dialog size when changing tabs:
image
image

There are a few more nits, but I think if you just rebase onto the `main` branch, the lint rules and build checks should catch them.

cnkeats:
> and I'm also not a fan of this jump in dialog size when changing tabs:

@vinceau The three of us are in agreement on this one. Nikki and I were unsure of the best way to keep a constant dialogue width and thought best to get your input. Any code suggestions for this would be appreciated - I tried my hand at setting various width settings already and never quite got what I was after.

The rest of the changes are pushed but I'll request another review after I fix the dialogue width

vinceau (changes requested):
Getting there! There are too many problems with the React component I'll just fix the rest up myself after this PR is merged. Just got a few more nits for now.

vinceau (approved):
LGTM! Thanks for fixing those up!

PR #381, opened 2023-06-08
Fix button color and add mock chat messages to mock Slippi service

vinceau:
### Description

This PR changes the following:
* fixes buttons being a weird colour after upgrading the `@mui/material` package
* mocks the chat message functions in mock Slippi service
* adds a login notice to the chat options settings page when a user is not logged in (previously it will get stuck on "Loading..." forever)
* replaces the "Loading..." text with a spinner in the Chat options page.

#### Before
image

image

#### After
image

image


NikhilNarayana (approved):
lgtm

probably not necessary to handle in the mock but the difference is that tier 1 can use up to 8 premium chat messages while tier 2/3 can use 16.

PR #384, opened 2023-07-03
Minimal integration of kysely db

agiera:


vinceau:
Thanks for your help in looking into Kysely! We've integrated it into the repo in #400. :)

PR #385, opened 2023-07-13
feat: update relevant settings from imported ini and add jukebox toggle

NikhilNarayana:
This will allow users to enable the jukebox from the launcher. Additionally i added syncing settings back from dolphin on close

## with dolphin >=3.2.0
image

### DIFF
--- a/src/renderer/containers/Settings/MeleeOptions.tsx
+++ b/src/renderer/containers/Settings/MeleeOptions.tsx
@@ -9,11 +9,13 @@ import FormControlLabel from "@mui/material/FormControlLabel";
 import Radio from "@mui/material/Radio";
 import RadioGroup from "@mui/material/RadioGroup";
 import React from "react";
+import { lt } from "semver";
 
+import { Toggle } from "@/components/FormInputs/Toggle";
 import { PathInput } from "@/components/PathInput";
 import { useDolphinStore } from "@/lib/dolphin/useDolphinStore";
 import { useIsoVerification } from "@/lib/hooks/useIsoVerification";
-import { useIsoPath, useLaunchMeleeOnPlay } from "@/lib/hooks/useSettings";
+import { useIsoPath, useJukeboxEnabled, useLaunchMeleeOnPlay } from "@/lib/hooks/useSettings";
 
 import { SettingItem } from "./SettingItem";
 
@@ -37,8 +39,13 @@ export const MeleeOptions = React.memo(() => {
   const isoValidity = useIsoVerification((state) => state.validity);
   const [isoPath, setIsoPath] = useIsoPath();
   const [launchMeleeOnPlay, setLaunchMelee] = useLaunchMeleeOnPlay();
+  const [jukeboxEnabled, setJukeboxEnabled] = useJukeboxEnabled();
   const netplayDolphinOpen = useDolphinStore((store) => store.netplayOpened);
   const playbackDolphinOpen = useDolphinStore((store) => store.playbackOpened);
+
+  const dolphinVersion = useDolphinStore((store) => store.netplayDolphinVersion);
+
+  const disableJukeboxToggle = dolphinVersion !== null && lt(dolphinVersion, "3.2.0");
 
   const onLaunchMeleeChange = async (value: string) => {
     const launchMelee = value === "true";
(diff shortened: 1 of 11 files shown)


src/renderer/components/FormInputs/Toggle.tsx
```diff
@@ -10,15 +11,25 @@ type ToggleProps = {
...
-export const Toggle = ({ label, description, value, onChange, disabled }: ToggleProps) => {
+export const Toggle = ({ label, description, value, onChange, disabled, tooltipTitle }: ToggleProps) => {
   return (
     <div>
       <FormControlLabel
         labelPlacement="start"
         control={
-          <Switch checked={value} onChange={(e) => onChange(e.target.checked)} color="primary" disabled={disabled} />
+          <Tooltip title={tooltipTitle}>
+            <div>
```
vinceau, line 24: Do we need this `div`? I assume you added it for a reason.

src/renderer/components/FormInputs/Toggle.tsx
```diff
@@ -10,15 +11,25 @@ type ToggleProps = {
   value: boolean;
   onChange: (val: boolean) => void;
   disabled?: boolean;
+  tooltipTitle?: string;
```
vinceau, line 14:
```suggestion
  tooltip?: string;
```
Maybe just rename this to just `tooltip`.

src/renderer/containers/Settings/MeleeOptions.tsx
```diff
@@ -79,6 +86,16 @@ export const MeleeOptions = React.memo(() => {
...
+      <SettingItem name="">
+        <Toggle
+          value={jukeboxEnabled}
+          onChange={(checked) => setJukeboxEnabled(checked)}
+          label="Enable Music"
+          description="Enable music in game via Slippi Jukebox. Incompatible with WASAPI."
+          disabled={disableJukeboxToggle || netplayDolphinOpen}
+          tooltipTitle={disableJukeboxToggle ? "This feature is not yet available" : undefined}
+        />
+      </SettingItem>
```
vinceau, line 98: I'm not a big fan of disabling a toggle that they cannot use. Can we just hide/show the toggle based on the Dolphin version?

src/renderer/lib/hooks/useSettings.ts
```diff
@@ -42,6 +42,14 @@ export const useMonthlySubfolders = () => {
   return [useMonthlySubfolders, setUseMonthlySubfolders] as const;
 };
 
+export const useJukeboxEnabled = () => {
```
vinceau, line 45:
```suggestion
export const useJukebox = () => {
```
I think this is cleaner.

vinceau (changes requested):
I think it's really confusing that the setting is called `jukeboxEnabled`. It should just be called jukebox and whether it's enabled or not should be determined by the value in the setting.

I also don't think we should have disabled toggles that users cannot change. I would prefer us to only show the toggle on supported Dolphin versions. This should mean the changes to the Toggle component can be reverted.

src/renderer/containers/Settings/MeleeOptions.tsx
NikhilNarayana: so i was on the fence about that, initially i set it up to show/hide but i wanted to give people an idea of the fact that it is coming, but that also might not be necessary

src/renderer/components/FormInputs/Toggle.tsx
NikhilNarayana: yeah the Tooltip won't work if it is wrapping a disabled component so you need a div in between

src/renderer/containers/Settings/MeleeOptions.tsx
vinceau: I think ideally this setting shouldn't even be necessary since it should just be normal for a game to have music. But I think it's probably necessary to have because of its incompatible with WASAPI which lot's of pro players use. I also do not think rolling out things that are "coming soon" should really be necessary either.

NikhilNarayana: > I think ideally this setting shouldn't even be necessary since it should just be normal for a game to have music.

nah, there has always been some method to turn off music

NikhilNarayana: I'll go ahead with the show/hide option

ryanmcgrath:
I would argue that it should be disable-able since it’s using resources in the background due to not being native game music. It’s certainly conceivable one might need to disable it.

I.e, this is a special case.

vinceau:
> since it’s using resources in the background due to not being native game music

Yeah I agree that it should be toggle-able because of the background resources + because it's incompatible with WASAPI. I'm more so advocating that ideally if the native game music did actually work, it would mean a user wouldn't have to actively think about another setting and enabling it.

vinceau (approved):

PR #388, opened 2023-08-15
feat: add betaNetplay and betaPlayback settings

NikhilNarayana:
the radio buttons disable if the user is not subbed or logged out because the backend only provides the beta download link if you're subbed. adding a tooltip to the radio buttons didn't look great cause it always ended up pointing from a weird spot. i'm looking for other ideas

beta dolphins are written to an adjacent folder to the stable dolphins to avoid any conflicts.

todo:
- update the settings sync to handle mainline properly

### User is subbed
image

### User is not subbed or logged out (tooltip version is just for reference, it is not in this PR atm)
image
image


src/dolphin/install/mainlineInstallation.ts
```diff
@@ -0,0 +1,298 @@
...
+import path from "path";
+import { lt } from "semver";
+import semverRegex from "semver-regex";
+
+import type { DolphinInstallation } from "../types";
+import { DolphinLaunchType } from "../types";
+import { downloadLatestDolphin } from "./download";
+import type { DolphinVersionResponse } from "./fetchLatestVersion";
+
+const log = electronLog.scope("dolphin/installation");
```
vinceau, line 18:
```suggestion
const log = electronLog.scope("dolphin/mainlineInstallation");
```
Do you think this would be helpful?

vinceau (approved):
LGTM with one suggestion.

PR #391, opened 2023-09-04
Add network diagnostic tool to advanced settings section

jmlee337:
This should make it easier for users and support volunteers to debug some network issues

screenshots: [https://imgur.com/a/4iNsmec](https://imgur.com/a/4iNsmec)

In the future, I intend to hook up the port mapping test to a slippi dolphin port mapping option (IE. press a button to turn on slippi dolphin port mapping if port mapping is available)


src/main/diagnostic.ts
```diff
@@ -0,0 +1,86 @@
+import type { PortMapping } from "@common/types";
+import { NatType, Presence } from "@common/types";
+import { createPmpClient, createUpnpClient } from "@xmcl/nat-api";
+import { gateway4async } from "default-gateway";
+import Tracer from "nodejs-traceroute-ts";
+import { createServer, request } from "stun";
+
+export async function natType(): Promise<{ address: string; natType: NatType }> {
```
NikhilNarayana, line 8:
```suggestion
export async function getNatType(): Promise<{ address: string; natType: NatType }> {
```

src/main/diagnostic.ts
```diff
@@ -0,0 +1,86 @@
...
+  const stunServer = createServer({ type: "udp4" });
+  const stunResponse1 = await request("stun1.l.google.com:19302", { server: stunServer });
+  const stunResponse2 = await request("stun2.l.google.com:19302", { server: stunServer });
+  const address1 = stunResponse1.getXorAddress();
+  const address2 = stunResponse2.getXorAddress();
+  stunServer.close();
+  return { address: address1.address, natType: address1.port === address2.port ? NatType.NORMAL : NatType.SYMMETRIC };
+}
+
+export async function portMapping(): Promise<PortMapping> {
```
NikhilNarayana, line 18:
```suggestion
export async function getPortMappingPresence(): Promise<PortMapping> {
```

src/main/diagnostic.ts
```diff
@@ -0,0 +1,86 @@
...
+    })
+    .finally(() => {
+      pmpClient.close();
+    });
+
+  await Promise.all([upnpPromise, pmpPromise]);
+  return { upnp: upnpPresence, natpmp: natpmpPresence } as PortMapping;
+}
+
+export async function cgnat(address: string): Promise<{ cgnat: Presence }> {
```
NikhilNarayana, line 62:
```suggestion
export async function getCgnatPresence(address: string): Promise<{ cgnat: Presence }> {
```

NikhilNarayana (changes requested):
i'm still in the process of testing functionality, but i got what i expected with no changes to my network

i would like the diagnostic option moved into the Help section since this doesn't really feel like an `Advanced App Setting`

src/renderer/containers/Settings/Diagnostic.tsx
```diff
@@ -0,0 +1,320 @@
...
+  return "NAT Type";
+};
+
+const getNatTypeCommentary = (natType: NatType) => {
+  switch (natType) {
+    case NatType.NORMAL:
+      return (
+        'Normal NAT - This is the optimal result. Also known as "easy", "full cone", "moderate", "open", ' +
+        '"port restricted cone", or "restricted cone" NAT.'
+      );
```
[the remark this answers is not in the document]
jmlee337: a bit confused here. wouldn't the suggested edit introduce a line break into the string?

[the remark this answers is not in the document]
jmlee337: discussed via discord - HTML elides whitespace so introducing line breaks doesn't matter here

src/renderer/containers/Settings/Diagnostic.tsx
```diff
@@ -0,0 +1,313 @@
...
+        <Typography variant="subtitle2">{portMappingTitle}</Typography>
+        <DialogBody>{portMappingDescription}</DialogBody>
+      </>
+    );
+
+  const cgnatTitle = getCgnatTitle(cgnat);
+  const cgnatDescription = getCgnatDescription(cgnat);
+  const tracerouteCommand = window.electron.common.isWindows ? "tracert" : "traceroute";
+  const cgnatCommand = tracerouteCommand + " " + ipAddress;
+  const displayedCgnatCommand = tracerouteCommand + " " + (ipAddressHidden ? hiddenIpAddress : ipAddress);
```
NikhilNarayana, line 226:
```suggestion
  const cgnatCommand = `${tracerouteCommand} ${ipAddress}`;
  const displayedCgnatCommand = `${tracerouteCommand} ${ipAddressHidden ? hiddenIpAddress : ipAddress}`;
```

src/main/api.ts
```diff
@@ -66,6 +69,18 @@ export default {
     const { result } = await ipc_showOpenDialog.renderer!.trigger(options);
     return result;
   },
+  async diagnosticNat() {
```
NikhilNarayana, line 72:
```suggestion
  async runNatDiagnostic() {
```
use this format for all of these, including the IPC calls

NikhilNarayana (dismissed):
let's move the Diagnostic component from Advanced App Settings to Help. put it right underneath `<SupportBox>` on line 21

src/renderer/containers/Settings/Diagnostic.tsx
```diff
@@ -0,0 +1,313 @@
...
+  const runDiagnostic = async () => {
+    const portMappingPromise = window.electron.common
+      .diagnosticPortMapping()
+      .then((resPortMapping) => {
+        setPortMapping(resPortMapping);
+      })
+      .catch(() => {
+        setPortMapping({ upnp: Presence.FAILED, natpmp: Presence.FAILED });
+      });
+    return Promise.all([runNatCgnatDiagnostics(), portMappingPromise]);
```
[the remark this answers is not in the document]
jmlee337: ```Promises must be awaited, end with a call to .catch, end with a call to .then with a rejection handler or be explicitly marked as ignored with the `void` operator.eslint[@typescript-eslint/no-floating-promises](https://typescript-eslint.io/rules/no-floating-promises)```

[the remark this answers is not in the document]
jmlee337: (changed to `await Promise.all([runNatCgnatDiagnostics(), portMappingPromise]);` sg?)

jmlee337:
> let's move the Diagnostic component from Advanced App Settings to Help. put it right underneath `<SupportBox>` on line 21

done: [https://imgur.com/a/bqCTVMB](https://imgur.com/a/bqCTVMB)

src/renderer/containers/Settings/Diagnostic.tsx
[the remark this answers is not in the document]
NikhilNarayana: ah whoops, i thought i had replaced return with await. that's exactly what i wanted, thanks

jmlee337:
looks nice! updated screenshot <https://imgur.com/a/aab8SU0>

package.json
```diff
@@ -77,6 +78,7 @@
     "@types/react-virtualized-auto-sizer": "^1.0.0",
     "@types/react-window": "^1.8.2",
     "@types/terser-webpack-plugin": "^5.0.4",
+    "@types/validator": "^13.11.1",
```
vinceau, line 81: Why do we import a typings for a package that we don't use?

src/main/setup.ts
```diff
@@ -186,4 +190,31 @@ export default function setupMainIpc({ dolphinManager }: { dolphinManager: Dolph
...
+  });
+
+  ipc_runDiagnosticCgnat.main!.handle(async ({ address }) => {
+    try {
+      const result = await getCgnatPresence(address);
+      return result;
+    } catch {
+      return { cgnat: Presence.FAILED };
+    }
+  });
```
vinceau, line 219: We don't need 3 different IPC endpoints when they're always called together. Just have a single IPC endpoint that fetches and returns all the necessary info required to render the FE.

src/renderer/containers/Settings/Diagnostic.tsx
```diff
@@ -0,0 +1,343 @@
...
+        Check network issues
+      </ActionButton>
+      <Dialog open={dialogOpen} closeAfterTransition={true} onClose={() => setDialogOpen(false)} fullWidth={true}>
+        <DialogTitle>Network Diagnostic</DialogTitle>
+        <DialogContent>
+          {getNatTypeSection()}
+          {getPortMappingSection()}
+          {getCgnatSection()}
+          {getCgnatCommandSection()}
+        </DialogContent>
```
vinceau, line 323:
```suggestion
        <DialogContent>
          <NatTypeSection natType={natType} />
          <PortMappingSection portMapping={portMapping} />
          <CgnatSection cgnat={cgnat} />
          <CgnatCommandSection whateverPropsAreNecessaryToRender={stuff} />
        </DialogContent>
```

This is a React component so prefer to compose individual components into one rather than having all these functions and strings everywhere.

src/renderer/containers/Settings/SupportBox.tsx
```diff
@@ -81,6 +83,9 @@ export const SupportBox = () => {
             Copy logs
           </Button>
         </div>
+        <div>
+          <Diagnostic />
```
vinceau, line 87:
```suggestion
          <NetworkDiagnosticsButton />
```
`Diagnostic` is very generic. This is specific to network diagnostics so it should be renamed as such.

src/renderer/containers/Settings/SupportBox.tsx
```diff
@@ -10,6 +10,8 @@ import { Button } from "@/components/FormInputs";
 import { useToasts } from "@/lib/hooks/useToasts";
 import { ReactComponent as DiscordIcon } from "@/styles/images/discord.svg";
 
+import { Diagnostic } from "./Diagnostic";
```
vinceau, line 13:
```suggestion
import { NetworkDiagnosticsButton } from "./NetworkDiagnosticsButton";
```

src/renderer/containers/Settings/Diagnostic.tsx
```diff
@@ -0,0 +1,343 @@
...
+    case Presence.FAILED:
+      return `Please try again later.
+        If the failure persists, you can test this in your computer's terminal app.
+        See below:`;
+    default:
+      return "";
+  }
+};
+
+export const Diagnostic = React.memo(() => {
```
vinceau, line 116:
```suggestion
export const NetworkDiagnosticsButton = React.memo(() => {
```

src/main/setup.ts
vinceau, line 219:
```suggestion
  ipc_runNetworkDiagnostics.main!.handle(async () => {
     // just do everything we need to do and return the final object here
     return {
       // Return all the different things here
     }
  });
```

src/main/diagnostic.ts
vinceau, line 1: This file does network diagnostics so we should call it `network_diagnostics.ts`

src/main/diagnostic.ts
```diff
@@ -0,0 +1,86 @@
+import type { PortMapping } from "@common/types";
+import { NatType, Presence } from "@common/types";
+import { createPmpClient, createUpnpClient } from "@xmcl/nat-api";
+import { gateway4async } from "default-gateway";
+import Tracer from "nodejs-traceroute";
+import { createServer, request } from "stun";
+
+export async function getNatType(): Promise<{ address: string; natType: NatType }> {
```
vinceau, line 8: If these functions are always called together we don't need to export all three functions. Just export a single function `getNetworkDiagnostics()` which calls all the other functions and returns everything that we need to render in the FE.

vinceau (changes requested):
Thanks for this PR. There are some changes I'd like to see before this gets merged.

src/main/diagnostic.ts
```diff
@@ -0,0 +1,86 @@
...
+  const stunServer = createServer({ type: "udp4" });
+  const stunResponse1 = await request("stun1.l.google.com:19302", { server: stunServer });
```
vinceau, line 10: Also can you have this as a constant?
```suggestion
  const stunResponse1 = await request(GOOGLE_STUN_SERVER, { server: stunServer });
```
where you also have:
```ts
const GOOGLE_STUN_SERVER = "stun1.l.google.com:19302";
```
ealier in the file, after the imports.

jmlee337:
> Thanks for this PR. There are some changes I'd like to see before this gets merged.

done!

src/renderer/containers/Settings/NetworkDiagnostic/CgnatCommandSection.tsx
```diff
@@ -0,0 +1,68 @@
...
+            {cgnatCommandCopied ? "Copied!" : "Copy"}
+          </Button>
+        </AlignCenterDiv>
+        <DialogBody>
+          {"More than one hop to your external IP address indicates CGNAT or Double NAT (or VPN)."}
+        </DialogBody>
+      </>
+    );
+  }
+  return <></>;
```
vinceau, line 67:
```suggestion
    return (
      <div>
        <Typography variant="subtitle2">Run this command</Typography>
        <AlignCenterDiv>
          <InputBase css={inputBaseCss} disabled={true} value={displayedCgnatCommand} />
          <Button variant="contained" color="secondary" onClick={onCgnatCommandShowHide} style={buttonStyle}>
            {cgnatCommandHidden ? "Reveal" : "Hide"}
          </Button>
          <Button variant="contained" color="secondary" onClick={onCgnatCommandCopy} style={buttonStyle}>
            {cgnatCommandCopied ? "Copied!" : "Copy"}
          </Button>
        </AlignCenterDiv>
        <DialogBody>
          More than one hop to your external IP address indicates CGNAT or Double NAT (or VPN).
        </DialogBody>
      </div>
    );
  }
```
I would prefer it if we use `<div>` over React fragments unless there's a very strong reason why we *need* to have the DOM children at the same level.
Additionally, I would prefer it if we leave the conditional rendering to the parent. e.g.
```ts
{(cgnat === Presence.FAILED && address) && <CgnatCommandSection />}
```

Can you also ensure that text contents in elements aren't unnecessarily wrapped in `{"` and `"}`? i.e. prefer `<div>text</div>` over `<div>{"text"}</div>`. This goes for all the `.tsx` files -- not just this one.

src/renderer/containers/Settings/NetworkDiagnostic/PortMappingSection.tsx
vinceau, line 1: This file is trivial enough to just be in-lined. Additionally, we shouldn't need to handle individual loading states since the fetching of the data for all sections happens at once. Therefore, we just need a single loading state, rather than each section handling their own loading state.

src/renderer/containers/Settings/NetworkDiagnostic/CgnatSection.tsx
vinceau, line 1: This file is also trivial enough to just be inlined. See other comment about not needing to handle loading state.

src/renderer/containers/Settings/NetworkDiagnostic/NetworkDiagnosticsButton.tsx
```diff
@@ -0,0 +1,186 @@
...
+  };
+
+  const openDialog = async () => {
+    setIpAddress("");
+    setNatType(NatType.UNKNOWN);
+    setPortMapping({ upnp: Presence.UNKNOWN, natpmp: Presence.UNKNOWN });
+    setCgnat(Presence.UNKNOWN);
+    setDialogOpen(true);
+    await runNetworkDiagnostics();
+  };
```
vinceau, line 114:
```suggestion
type NetworkInformation = {
  address: string,
  cgnat: Presence,
  natType: NatType,
  portMapping: PortMapping,
}

export const NetworkDiagnosticsButton = React.memo(() => {
  const [dialogOpen, setDialogOpen] = React.useState(false);
  const [isLoading, setIsLoading] = React.useState(false);
  const [isError, setIsError] = React.useState(false);
  const [networkInfo, setNetworkInfo] = React.useState<NetworkInformation | undefined>(undefined);
  
  const runNetworkDiagnostics = async () => {
    setIsLoading(true);
    setIsError(false);
    await window.electron.common.runNetworkDiagnostics().then(setNetworkInfo).catch(() => {
      // Probably want to handle this error state somehow with an error alert.
      setIsError(true);
    }).finally(() => setIsLoading(false));
  };

  const openDialog = async () => {
    setDialogOpen(true);
    await runNetworkDiagnostics();
  };
```

You're conflating the loading state with unknown network type. We should be handling the loading state (and error state technically) explicitly. Since all the data is being fetched at once, again we do not need individual loading states for each section (unless they're doing their own async calls).

vinceau (changes requested):
Getting there!

I've noticed a few common things:
1. Prefer returning a `<div>` over returning react fragments, unless the structure of the DOM is important enough to require you to.
2. Prefer returning `null` over an empty react fragment `<></>`.
3. Prefer inline text content `<div>text</div>` over wrapping in a string literal `<div>{"text"}</div>`

jmlee337:
> Getting there!
> 
> I've noticed a few common things:
> 
> 1. Prefer returning a `<div>` over returning react fragments, unless the structure of the DOM is important enough to require you to.
> 2. Prefer returning `null` over an empty react fragment `<></>`.
> 3. Prefer inline text content `<div>text</div>` over wrapping in a string literal `<div>{"text"}</div>`

all comments addressed!

vinceau (approved):

PR #392, opened 2023-09-23
ci: fix windows code signing

NikhilNarayana:
the gist of it is that signing keys now need to be stored in HSMs which means we won't have the signing cert available for electron-builder to do its thing. instead we have a custom script for signing that will handle calling CodeSignTool for each file that needs to be signed.
this will only sign release commits because we have to pay a service to manage our HSM. i hate this whole situation

### DIFF
--- a/.erb/scripts/windows-sign.js
+++ b/.erb/scripts/windows-sign.js
@@ -11,11 +11,17 @@ exports.default = async (config) => {
   const CODESIGNTOOL_PATH = process.env.CODESIGNTOOL_PATH;
   const fileToSign = config.path ? String(config.path) : "";
 
+  if (!ES_USERNAME) {
+    // we're probably not trying to sign if this isn't set so just exit safely
+    console.log("not signing binaries");
+    return;
+  }
+
   if (process.platform !== "win32") {
     throw new Error(`unexpected platform: ${process.platform}`);
   }
 
-  if ([ES_CREDENTIAL_ID, ES_USERNAME, ES_PASSWORD, ES_TOTP_SECRET, CODESIGNTOOL_PATH, fileToSign].some((v) => !v)) {
+  if ([ES_CREDENTIAL_ID, ES_PASSWORD, ES_TOTP_SECRET, CODESIGNTOOL_PATH, fileToSign].some((v) => !v)) {
     throw new Error("missing required secrets");
   }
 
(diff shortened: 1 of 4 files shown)


vinceau (approved):
LGTM once fixes are made but can you update the description of this PR with more information and context for this change? Without the context, it looks like there's a lot of changes going on to just "only sign release commits".

.erb/scripts/windows-sign.js
```diff
@@ -0,0 +1,32 @@
...
+  const ES_TOTP_SECRET = process.env.ES_TOTP_SECRET;
+  const CODESIGNTOOL_PATH = process.env.CODESIGNTOOL_PATH;
+  const fileToSign = config.path ? String(config.path) : "";
+
+  if (
+    process.platform !== "win32" ||
+    [ES_CREDENTIAL_ID, ES_USERNAME, ES_PASSWORD, ES_TOTP_SECRET, CODESIGNTOOL_PATH, fileToSign].some((v) => !v)
+  ) {
+    console.log("missing something");
+    return;
```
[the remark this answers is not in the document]
NikhilNarayana: yeah you're right. updated

.erb/scripts/windows-sign.js
```diff
@@ -0,0 +1,31 @@
...
+  const CODESIGNTOOL_PATH = process.env.CODESIGNTOOL_PATH;
+  const fileToSign = config.path ? String(config.path) : "";
+
+  if (
+    process.platform !== "win32" ||
+    [ES_CREDENTIAL_ID, ES_USERNAME, ES_PASSWORD, ES_TOTP_SECRET, CODESIGNTOOL_PATH, fileToSign].some((v) => !v)
+  ) {
+    throw new Error("missing required secrets");
+  }
+
```
vinceau, line 20:
```suggestion
  if (process.platform !== "win32") {
    throw new Error(`unexpected platform: ${process.platform}`);
  }
  
  if ([ES_CREDENTIAL_ID, ES_USERNAME, ES_PASSWORD, ES_TOTP_SECRET, CODESIGNTOOL_PATH, fileToSign].some((v) => !v)
  ) {
    throw new Error("missing required secrets");
  }

```

vinceau (commented):
I think we should also split these two conditionals so we know exactly why the error is thrown.

.erb/scripts/windows-sign.js
```diff
@@ -0,0 +1,37 @@
...
+// Custom Sign hook for use with electron-builder + SSL CodeSignTool
+
+const { execSync } = require("child_process");
+
+exports.default = async (config) => {
+  const ES_CREDENTIAL_ID = process.env.ES_CREDENTIAL_ID;
+  const ES_USERNAME = process.env.ES_USERNAME;
+  const ES_PASSWORD = process.env.ES_PASSWORD;
+  const ES_TOTP_SECRET = process.env.ES_TOTP_SECRET;
+  const CODESIGNTOOL_PATH = process.env.CODESIGNTOOL_PATH;
```
vinceau, line 11: Do you mind also moving these outside/above this function, so they appear more like actual constants?

PR #395, opened 2023-10-07
Add config flags

vinceau:
### Description

This PR adds build-time config flags, as well as runtime flags which can override the build flags.

For now we only support the following two flags:

```bash
# When building the app, set in your environment to enable
ENABLE_MAINLINE_DOLPHIN=1  # --enable-mainline-dolphin
ENABLE_REPLAY_DATABASE=1  # --enable-replay-database
```


src/main/flags/flags.ts
```diff
@@ -0,0 +1,74 @@
...
+    enableReplayDatabase: parseBoolean(process.env.ENABLE_REPLAY_DATABASE, false),
+  };
+}
+
+function getRuntimeFlags(): Partial<ConfigFlags> {
+  const overrides: Partial<ConfigFlags> = {};
+  const args = process.argv.slice(1);
+
+  let i = 0;
+  while (i < args.length) {
```
vinceau, line 43: We manually define the index outside the while loop so that in the future, if we want to handle flags which can accept an argument e.g. `--some-flag value` we can manually increment `i` by however many args we want to consume.
If we instead decide to use the format of `--some-flag=value` then we can remove this, but until then this keeps things flexible.

NikhilNarayana (approved):

PR #400, opened 2023-10-28
Add DatabaseReplayProvider

vinceau:
### Description

This PR aims to finally address #178 by adding the foundation for a database in SQLite3 to store replay information. It uses the implementation `better-sqlite3` and the query builder `kysely`. This PR has been made possible by the contributors of the past works in #364 #383 #384. Many thanks to those involved.

#### Run jest via Electron

Since `better-sqlite3` is a native node module, we need to update the command used to run tests. See [this comment](https://github.com/WiseLibs/better-sqlite3/issues/545#issuecomment-824887942) to know why. This allows us to run the jest tests with the same node version as Electron.

#### Migrations Handling

We need to package all schema migrations into the package itself for it to be run by `kysely` on the user's machine. To do this:
1. We compile the Typescript migrations into Javascript with the new `yarn run build:migrations` command, with the process defined in `webpack.config.migrations.prod.ts`
2. We copy the `release/app/dist/migrations/*.js` files into the [extraResources](https://www.electron.build/configuration/contents.html#extraresources) folder, see `electron-builder.json`
3. If the app is packaged, we execute the migrations from the `extraResources/migrations` folder, otherwise we run it directly from the `src/database/migrations` folder. See the `create_database.ts` file for the implementation.

### Remaining tasks
- [x] Better packaging of migration files
- [x] Somehow move all database handling into a worker
- [x] Infer game start time *before* inserting into the database, rather than *after*
- [x] Add more fields like set info, and unique player id
- [x] Add option to re-index the whole DB from scratch. We likely want this when we make major changes to the schema.


src/replays/database_replay_provider/database_replay_provider.ts
```diff
@@ -0,0 +1,268 @@
...
+      FileRepository.findTotalSizeByFolder(this.db, folder),
+    ]);
+
+    const players = await Promise.all(
+      gameRecords.map(async ({ _id: gameId }): Promise<[number, PlayerRecord[]]> => {
+        const playerRecords = await PlayerRepository.findAllPlayersByGame(this.db, gameId);
+        return [gameId, playerRecords];
+      }),
+    );
+    const playerMap = new Map(players);
```
alvarosevilla95, line 70: This seems a bit slow on a relatively large db.

alvarosevilla95 (commented):
Hey @vinceau, been a while! Awesome to see this getting added. 

The PR looks fantastic tbh. I really like all the decisions on the schema and db manteinance. I wasn't familiar with kysely before, but it looks great honestly. The typings and the query builders are cool, and so is the migration support it seems. Not much to comment on the schema itself, it looks very reasonable (although you know I'd love to eventually add game stats to the schema :D), and the indexing happens very fast even on a large folder.

I only noticed one small slow point. Fetching players by gameId one by one takes about 4 seconds on my 10k games, which is visible when clicking Refresh. Changing the code to do the join in the db using 'folder' speeds it up to 200ms. I pushed a commit to https://github.com/alvarosevilla95/slippi-desktop-app/commit/44f0972612849ab6b0a8f0d963e59dd15ceccfc0 if you want to cherry-pick

Great stuff, looking forward to seeing this live!

NikhilNarayana (approved):

src/replays/database_replay_provider/database_replay_provider.ts
vinceau: Fixed in 43141c38. :)

PR #411, opened 2023-12-29
Introduce StyleX for styling

vinceau:
### Description

This PR introduces yet another styling option into our codebase lol. This time it's Facebook's StyleX.

### Motivation

Currently we have a lot of different styling options in our codebase:
1. Emotion
2. material-ui style overrides via `sx`
3. inline style props
4. plain ol' css

And then even with Emotion, there are different ways to style components e.g.

```tsx
const SomeComponent = styled.div`
  border: soild 2px red;
`;
```

or:

```tsx
const SomeComponent = () => {
  return (<div css={css`border: solid 2px red`}/>)
};
```

As a developer it's really not clear what we should be using when and where. I'd be nice if we could reduce this list down to just one consistent method we use to style components across the board.

#### What's wrong with Emotion?

Defining the Emotion styles as string literals, the entire CSS strings are included as is in the compiled JS blob. There are a few issues with this:

1. There aren't any smarts to deduplicate styles.
2. There is no syntax highlighting or syntax checking since it's just a string, anything goes.
3. Increases the size of our `renderer.js` and means there's more overhead in running and applying the styles.
4. Applying conditional logic is not straight-forward, and you end up with a lot of [messy conditional logic inside of the string](https://github.com/project-slippi/slippi-launcher/blob/1e0eef0416f6ce71ec3eb30dc5b0f1cd36d2ac9d/src/renderer/pages/replays/replay_browser/replay_file/replay_file.tsx#L111-L140)

#### What does StyleX offer over Emotion?

1. Build-time CSS generation

This is actually huge for performance because all the CSS is generated beforehand at build, so that at runtime it only needs to add/remove classnames. We should see a much better performing application if we switch to StyleX, ~and even more if we eventually add lazy loading to our pages too~ (addressed in #412).

2. Type-safety

Since it uses object notation (similar to the style props and Material's overrides), rather than strings for defining types, we're much less likely to introduce bugs.

### Migration Plan

Ideally we start using StyleX on all new components and modifications to existing components. Over time we reduce dependence on Emotion and switch fully over to StyleX for all custom components and styling needs.

**End goal**: Unfortunately we cannot completely remove Emotion from our codebase since it is a dependency of Material UI. However we can at least stop using it explicitly except for MUI themes and overrides.


NikhilNarayana (approved):
seems like a solid improvement to me

PR #412, opened 2024-01-04
Lazy load pages

vinceau:
Ths PR refactors a bunch of the renderer to support lazy loading pages on demand. This should hopefully make our app load a bit faster, and should make it easier to traverse the project directory.

Pages are created using a `create.tsx` file, and a `load.tsx` will import the `create.tsx` file on demand when the page needs to be rendered.

New directory structure:
* `app`: Basically what the `views` folder was. Acts as an entry point for the app.
* `components`: This will be only for shared components. These should still be dumb-components.
* `containers`: Hopefully we can delete this folder eventually, but for now there are still some shared containers that make it unclear where these components should live.
* `pages`: Basically kind of what containers were but more likely to match the actual pages of the app. Keeping the components/containers that it uses in its own sub-folder. 

We also take this refactor to convert all file names in `renderer` to be `snake_case`. This should add some consistency to the directory structure.


NikhilNarayana (approved):
> hopefully make our app load a bit faster

my measurements (taking videos and counting frames) showed only a marginal improvement in load times, but that's better than huge losses
﻿
i like the reorganization, it is a lot easier to know where to start drilling down from

PR #450, opened 2024-12-29
Add player profile links to the replay browser

juebjueb:
Follow up on PR #375 authored by [trstruth](https://github.com/project-slippi/slippi-launcher/pull/375#issue-1616061627).

image

Addressing comments from [vinceau](https://github.com/project-slippi/slippi-launcher/pull/375#issuecomment-1462981675):
> 1. The inconsistent behaviour of some badges being clickable and some not is quite jarring.
> 2. It occupies the badge click functionality in case we wanted to use that behaviour for something else in the future.
> 3. The `PlayerBadge` which should be a dumb component now also has to handle the click function which is undesirable. I'd rather leave that to the parent.

1. Added an explicit button to eliminate confusion for which badges are clickable.
2. I agree that a pop-up menu would be nice to allow further extension, but I worry this would be premature. There hasn't been much activity on the replay browser over the last year, and I wouldn't want there to be a menu that's only populated with one item for a long period of time.
3. Moved from PlayerBadge to its parent TeamElements

### DIFF
--- a/src/renderer/pages/replays/replay_file_stats/player_info.tsx
+++ b/src/renderer/pages/replays/replay_file_stats/player_info.tsx
@@ -1,6 +1,8 @@
 import { css } from "@emotion/react";
 import styled from "@emotion/styled";
+import Tooltip from "@mui/material/Tooltip";
 
+import { ExternalLink as A } from "@/components/external_link";
 import { getColor } from "@/lib/player_colors";
 import { getCharacterIcon } from "@/lib/utils";
 import { colors } from "@/styles/colors";
@@ -32,6 +34,7 @@ export const PlayerInfo = ({
   const port = playerIndex + 1;
   const backupName = type === 1 ? "CPU" : `Player ${port}`;
   const charIcon = getCharacterIcon(characterId ?? null, characterColor);
+  const slippiProfileUrl = `https://slippi.gg/user/${connectCode?.split("#").join("-")}`;
   return (
     <Outer>
       <div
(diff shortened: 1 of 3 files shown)


erickaoek:
This is awesome, someone review please :)

JLaferri:
I think this is cool and would be great to merge but I think the visual styling needs a little work. It's not awful but it certainly feels tacked on and like it doesn't fully belong. I'm not sure what my suggestion is to fix it atm though. I'm sure a designer could come up with a clean solution.

juebjueb:
Probably the visual issue is that the links are in between the player badges and the "vs". How about moving to the replay browser like so? Adds an extra click to get there, but looks nicer.

image

JLaferri:
> Probably the visual issue is that the links are in between the player badges and the "vs". How about moving to the replay browser like so? Adds an extra click to get there, but looks nicer.

I think I would personally be okay with that for now.

juebjueb:
Great, thanks! Ready for review.

vinceau:
what about moving the link to the actual connect code at the top of the replay stats page banner?

juebjueb:
We could, but this would have the same visual jank as the before where the links are in between the player names and the VS which breaks symmetry

vinceau:
> We could, but this would have the same visual jank as the before where the links are in between the player names and the VS which breaks symmetry

by links and visual symmetry are you referring to the external link visual indicator? because i was imagining not having that visual indicator and just having like an underline or something to indicate interactivity and clickable link. i don't see how that would break symmetry or have visual jank

juebjueb:
Done, see latest.

image

src/renderer/pages/replays/replay_file_stats/player_info.tsx
```diff
@@ -85,7 +87,16 @@ export const PlayerInfo = ({
...
-            {connectCode}
+            <A
+              css={css`
+                &:hover {
+                  opacity: 0.7;
+                }
+              `}
+              href={slippiProfileUrl}
+            >
+              <div>{connectCode}</div>
```
vinceau, line 98:
```suggestion
              {connectCode}
```
can you remove the nested div here? we shouldn't need it.

vinceau (changes requested):

juebjueb:
done

PR #458, opened 2025-07-20
Reducing min window sizes

NikAgs:
Please I just want to make it smaller so I can have other windows open without overlap

I checked the entire app is responsive at min size still. 

Screenshot 2025-07-19 221151

### DIFF
--- a/src/main/main.ts
+++ b/src/main/main.ts
@@ -103,8 +103,8 @@ const createWindow = async () => {
     show: false,
     width: 1100,
     height: 728,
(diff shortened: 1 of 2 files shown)


NikAgs:
@NikhilNarayana @vinceau

NikhilNarayana:
provide some justification for why this change is good/necessary. i currently don't see a reason to make it so small

JLaferri:
This is not something we will merge. We sometimes show important information about downtime or otherwise in the main section of the launcher and allowing people to hide that easily is not a good idea.

PR #461, opened 2025-11-05
Add i18nService and support for changing app language

vinceau:
### Description

This PR adds some initial support for an i18n service with preliminary Spanish and Japanese translation for some strings.

We use the [i18next-auto-keys plugin](https://www.npmjs.com/package/i18next-auto-keys) to extract keys and generate PO files, and also to convert them back into JSON.

This PR adds:
* an i18nService
* a LanguageSelector dropdown component
* preliminary support for ES and JA translations

I say preliminary support beecause I think this PR is a good foundation for i18n and should allow for contributors to easily add new languages and help translate missing strings. Therefore this PR should be merged even without 100% string translation. English fallback is default for missing translations.

### Screenshots
Screenshot 2025-11-07 at 1 05 52 AM

Screenshot 2025-11-07 at 1 06 15 AM

Screenshot 2025-11-07 at 1 06 28 AM

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -4,14 +4,18 @@ Thank you for your interest in contributing to the Slippi Launcher! This documen
 
 ## Adding Translations
 
-The Slippi Launcher uses a custom internationalization (i18n) system built on top of `i18next` with the `i18next-auto-keys` package. This approach provides several benefits over traditional React i18n libraries.
+The Slippi Launcher uses i18next with the [i18next-auto-keys](https://www.npmjs.com/package/i18next-auto-keys) package for its translations system. This approach provides several benefits over traditional React i18n libraries.
 
 ### Benefits Over react-i18next
 
-1. **No Component Pollution**: Unlike `react-i18next`, our system doesn't require wrapping components with translation providers or importing translation hooks throughout the codebase
-2. **Build-Time Optimization**: Messages are transformed at build time, resulting in better performance and smaller bundle sizes
-3. **Cleaner Code**: Components remain focused on their logic without translation concerns scattered throughout
-4. **Automatic Key Generation**: No need to manually manage translation keys - they're generated automatically from your message functions
+1. **No Component Pollution**: Doesn't require wrapping components with translation providers or importing translation hooks throughout the codebase. The developer doesn't even need to know what the translation system under the hood is.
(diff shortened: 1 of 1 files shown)


locales/es.po
```diff
@@ -0,0 +1,252 @@
...
+msgid "Clear files"
+msgstr "Limpiar archivos"
+
+#: src/renderer/pages/settings/game_settings/game_settings.messages.ts:4:31
+msgid "Close Dolphin to change this setting"
+msgstr "Cierra Dolphin para cambiar esta configuración"
+
+#: src/renderer/app/create.messages.ts:5:18
+msgid "Console Mirror"
+msgstr "Espejo de Consola"
```
nachoverdon: I think that Console Mirror should not be translated here, at least not the mirror word. I've never heard the expression "Espejo de consola" ever. I'd suggest against translating the term at all and use `"Mirror de Consola"` or use another term like console transmission/syncronization or maybe console cloning/duplication, tho these last two sound like you are cloning the console itself lol

Those would be:
- `"Transmisión de Consola"`
- `"Sincronización de Consola"`
- `"Duplicación de Consola"`
- `"Clonación de Consola"`

locales/es.po
```diff
@@ -0,0 +1,252 @@
...
+#. {0} count: number
+msgid "{0, plural, one {# file} other {# files}} found."
+msgstr "{0, plural, one {# archivo encontrado} other {# archivos encontrados}}."
+
+#: src/renderer/pages/replays/replay_browser/replay_browser.messages.ts:4:38
+#. {0} count: number
+msgid "{0, plural, one {# file} other {# files}} had errors."
+msgstr ""
+"{0, plural, one {# archivo tenía errores} other {# archivos tenían "
+"errores}}."
```
nachoverdon: This could just be `"{0, plural, one {# archivo tenía} other {# archivos tenían}} errores."`, as the word "errores" doesn't change

Also, I'm not sure why there's a a empty string at the beginning there, don't really know if necessary given that the original doesn't have it

JLaferri (approved):

locales/es.po
vinceau: Fixed! The empty strings just indicate that the string is split across multiple lines. It doesn't make a functional difference if it's split across multiple lines or not.

locales/es.po
vinceau: Went with the first suggestion

PR #468, opened 2025-11-22
Spectate Remote Control resub bugfix + minor api additions

jmlee337:
- make sure we re-sub to spectate controller when stopping and starting the remote control server
- add `broadcastId` to `spectate-broadcast-response`, `new-file-event`, and `game-end-event`
- spectateDetails `playbackId` -> `dolphinId` for consistency


vinceau (approved):
LGTM. I assume you've tested this change?

PR #480, opened 2025-11-27
fix: clean up unterminated workers

vinceau:
### Description

This PR cleans up the workers and avoids memory leaks.

### Context

We create the workers on demand. That's fine. But they're never disposed except on app close. They should be appropriately disposed when not needed.

We have three workers:
* Broadcast worker -> Should dispose when a user stops broadcasting
* Mirror worker -> Should dispose when there are no more connected consoles
* Spectate worker -> This one is trickier since we don't really know when exactly we should terminate and close our connection with the spectate server. I've set it to set a 15 minute timeout when no more spectate activity occurs.

### DIFF
--- a/src/broadcast/setup.ts
+++ b/src/broadcast/setup.ts
@@ -35,6 +35,18 @@ export default function setupBroadcastIpc({
   let prefixOrdinal = 0;
   let spectateIdleTimer: NodeJS.Timeout | undefined;
 
+  const disconnectFromSpectateServerIfInactive = async () => {
+    if (spectateWorker) {
+      const openBroadcasts = await spectateWorker.getOpenBroadcasts().catch(() => []);
+      if (openBroadcasts.length === 0) {
+        log.debug("Disconnecting and terminating worker due to idle timeout");
+        await spectateWorker.disconnect();
+        await spectateWorker.terminate();
+        spectateWorker = undefined;
+      }
+    }
+  };
+
   // Helper to start/reset the idle timeout (only when no active broadcasts)
   const resetSpectateIdleTimeout = async () => {
     // Clear any existing timer
@@ -48,14 +60,7 @@ export default function setupBroadcastIpc({
       const openBroadcasts = await spectateWorker.getOpenBroadcasts().catch(() => []);
       if (openBroadcasts.length === 0) {
         log.debug("Starting spectate idle timeout");
-        spectateIdleTimer = setTimeout(async () => {
-          log.debug("Spectate idle timeout reached, disconnecting and terminating worker");
-          if (spectateWorker) {
-            await spectateWorker.disconnect();
-            await spectateWorker.terminate();
-            spectateWorker = undefined;
-          }
-        }, SPECTATE_IDLE_TIMEOUT);
+        spectateIdleTimer = setTimeout(disconnectFromSpectateServerIfInactive, SPECTATE_IDLE_TIMEOUT);
       }
     }
   };


NikhilNarayana (approved):

src/broadcast/setup.ts
```diff
@@ -56,6 +106,10 @@ export default function setupBroadcastIpc({
     );
 
     await spectateWorker.refreshBroadcastList();
+
+    // Reset idle timeout on refresh
+    await resetSpectateIdleTimeout();
```
jmlee337: only reset spectate idle timeout if idle timeout has been started. It may have been stopped below because we started watching a broadcast, in which case we don't want to reset (implicitly start) it here

src/broadcast/setup.ts
```diff
@@ -46,6 +92,10 @@ export default function setupBroadcastIpc({
       spectateWorker = await createSpectateWorker(dolphinManager);
     }
     await spectateWorker.connect(authToken);
+
+    // Reset idle timeout on connect
+    await resetSpectateIdleTimeout();
```
jmlee337: only reset (implicitly start) spectate idle timeout if we weren't already connected. We can pipe through a return value from `SpectateManager` `connect()` to `SpectateWorker` `connect()` to here to indicate whether we were already connected.

src/broadcast/spectate_manager.ts
jmlee337: as of #490 it's possible we may have stopped watching a broadcast without closing the playback dolphin instance. We will need to create and emit a new `SpectateEvent` here in order to check if we should start the spectate idle timeout in this case

src/broadcast/setup.ts
```diff
@@ -30,14 +33,57 @@ export default function setupBroadcastIpc({
...
     })
-    .subscribe((event) => {
+    .subscribe(async (event) => {
       if (spectateWorker) {
-        void spectateWorker.dolphinClosed(event.instanceId).catch(log.error);
+        await spectateWorker.dolphinClosed(event.instanceId).catch(log.error);
+
+        // Check if there are any remaining open broadcasts
+        const openBroadcasts = await spectateWorker.getOpenBroadcasts().catch(() => []);
+        if (openBroadcasts.length === 0) {
```
jmlee337: I believe this check is redundant as `resetSpectateIdleTimeout()` already does it?

src/broadcast/setup.ts
```diff
@@ -30,14 +33,57 @@ export default function setupBroadcastIpc({
...
+  const resetSpectateIdleTimeout = async () => {
+    // Clear any existing timer
+    if (spectateIdleTimer) {
+      clearTimeout(spectateIdleTimer);
+      spectateIdleTimer = undefined;
+    }
+
+    // Only start timeout if worker exists and has no active broadcasts
+    if (spectateWorker) {
+      const openBroadcasts = await spectateWorker.getOpenBroadcasts().catch(() => []);
```
jmlee337: if my understanding of the event loop is correct, this should be currently safe from race conditions ([after requested changes] reading and mutating the underlying `openBroadcasts` `Record` only happens in synchronous functions), but a future change to make `SpectateManager` `watchBroadcast` async could unknowingly introduce that risk.

therefore I think it would be prudent to add a lock betwteen `SpectateManager` `watchBroadcast` and `getOpenBroadcasts`

src/broadcast/spectate_manager.ts
jmlee337: not sure what's going on with the comment here as there's no logic to disconnect from dolphin present

jmlee337: we should clean up `openBroadcasts` and `availableBroadcasts` here as in `disconnect()`

src/broadcast/spectate_manager.ts
```diff
@@ -304,4 +304,25 @@ export class SpectateManager extends EventEmitter {
       dolphinId: value.dolphinId,
     }));
   }
+
+  public async disconnect(): Promise<void> {
```
jmlee337: remove `async/Promise` as this function is not async

src/broadcast/spectate_manager.ts
```diff
@@ -304,4 +304,25 @@ export class SpectateManager extends EventEmitter {
...
+    this.emit(SpectateEvent.BROADCAST_LIST_UPDATE, []);
+    this.emit(SpectateEvent.LOG, "Disconnecting from spectate server");
+
+    // Stop watching all broadcasts
```
jmlee337: since this code path is currently only used to disconnect while idle, should we check `openBroadcasts` first and not continue if there are indeed open broadcasts?

jmlee337 (commented):
there seem to be a lot of gotchas doing spectate idle timeout. It may be more practical to simply add a button to "start/stop spectate" if we feel like that's good enough for broadcast.

PR #482, opened 2025-11-30
refactor: simplify quick start step generation

vinceau:
### Description

The current quick start logic is a bit convoluted with logic scattered everywhere. We refactor it to have a clear way to generate the steps once on app initialization, and a clear way to show the current step, and a clear way to redirect/navigate back to main if we don't have any steps that need completing.


src/renderer/components/login_form/login_form.tsx
```diff
@@ -34,13 +33,18 @@ const useLoginStore = create(
   ),
 );
 
+const Header = styled.h2`
```
NikhilNarayana: maybe we try to use stylex wherever we're creating new styles?

PR #485, opened 2025-12-01
debug: figure out why some tests are failing on ubuntu only

vinceau:
WIP - trying to debug why the ubuntu tests are failing


vinceau:
fixed in #488. we thought it was linux only but it's actually cuz we're only running the tests on the linux CI build lol.

PR #487, opened 2025-12-02
debug: linux database test 3

vinceau:
i will mald if the database tests fail


vinceau:
fixed in #488. we thought it was linux only but it's actually cuz we're only running the tests on the linux CI build lol.

PR #490, opened 2025-12-05
spectate remote `stop-broadcast-request`

jmlee337:


src/remote/spectate_remote_server.ts
```diff
@@ -228,6 +228,20 @@ export class SpectateRemoteServer {
         const err = typeof e === "string" ? e : e instanceof Error ? e.message : "unknown";
         this.connection.sendUTF(JSON.stringify({ op: "spectate-broadcast-response", err }));
       }
+    } else if (json.op === "stop-broadcast-request") {
```
NikhilNarayana: i think we should convert this to switch case with an enum for possible values and an error message for unknown op codes

NikhilNarayana (approved):

PR #506, opened 2025-12-12
i18n: add pt-BR translations

CaioIcy:
# Description

This PR adds the pt-BR translations (partially machine generated, reviewed and modified by me).

Small note, but it seems [this code](https://github.com/project-slippi/slippi-launcher/blob/d24e63cf349fb32dee73c38d3420b121e3ef28c5/src/renderer/services/i18n/i18n.service.ts#L48) prevents locale suffixes from being handled so I can't add the language as `pt-BR`.

```
Translation Progress Summary
══════════════════════════════════════════════════
✅ pt           [████████████████████] 100%
    File: pt.po
    Translated: 413/413 (0 remaining)
```

### DIFF
--- a/locales/pt.po
+++ b/locales/pt.po
@@ -7,7 +7,7 @@ msgstr ""
 "Plural-Forms: nplurals=2; plural=(n != 1);\n"
 "x-generator: i18next-auto-keys CLI\n"
 "Language: pt-BR\n"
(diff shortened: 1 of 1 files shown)


vinceau:
Our underlying translation library (i18next) has trouble resolving translations when it's using the full locale (believe me - I've tried to get it to resolve and it overcomplicates the entire setup), so having it just be pt.po is fine. Thank you for your contribution!

PR #523, opened 2026-01-30
fix: remove redundant multi-account sync effect

Raphie10:
**Summary**

Removes a redundant useEffect in UserMenu that manually synchronized multi-account state.

**Background**

Multi-account state (accounts and activeAccountId) is already kept in sync globally via
multiAccountService.onAccountsChange, which updates the useAccount store whenever the
service changes.
The additional sync performed in UserMenu duplicated this behavior.

**Issue**

In development (React StrictMode), the extra effect caused a feedback loop between:
- the component-level useEffect
- the global onAccountsChange subscription
- Zustand store updates
This resulted in a Maximum update depth exceeded crash.

**Fix**

Remove the redundant component-level sync

Rely solely on the existing global subscription for multi-account state updates in
src/renderer/listeners/install_app_listeners.ts

### DIFF
--- a/src/renderer/app/header/user_menu/user_menu.tsx
+++ b/src/renderer/app/header/user_menu/user_menu.tsx
@@ -67,14 +67,6 @@ export const UserMenu = ({ user, handleError }: { user: AuthUser; handleError: (
   // Get multi-account service
   const multiAccountService = authService.getMultiAccountService();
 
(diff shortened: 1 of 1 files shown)


vinceau (approved):
LGTM! Thanks for the fix!

PR #526, opened 2026-02-01
feat(mock-auth): add multi-account mock support for dev mode

Raphie10:
**Summary**

Adding multi-account support to the mock auth service
Allowing multiple predefined mock users (for example test, admin)
Enabling account switching without Firebase or network access
Keeping existing auth interfaces unchanged

**How to test**

Run the app in dev mode
Log in with:
test / test
admin / admin
Switch between accounts in the UI
Verify the active user updates correctly

**Notes**

This affects mock-only code paths
No production authentication behavior is changed

### DIFF
--- a/src/renderer/services/auth/multi_account.service.mock.ts
+++ b/src/renderer/services/auth/multi_account.service.mock.ts
@@ -66,7 +66,6 @@ class MockMultiAccountClient implements MultiAccountService {
 
     try {
       this._notifyAccountsChanged();
-
       this._initialized = true;
       log.info("Multi-account service initialized");
     } catch (err) {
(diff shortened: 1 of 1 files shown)


src/renderer/services/auth/auth.service.mock.ts
```diff
@@ -84,63 +83,56 @@ class MockAuthClient implements AuthService {
...
+    // Notify listeners
+    this._userSubject.next(this.getCurrentUser());
   }
 
   @delayAndMaybeError(SHOULD_ERROR)
   public async refreshUser(): Promise<void> {
-    this._updateCurrentUser({ emailVerified: true });
+    const auth = this._multiAccountService.getActiveAuth();
+    Preconditions.checkExists(auth, "No active account");
+    Preconditions.checkExists(auth.currentUser, "User is not logged in.");
```
vinceau: Likewise as above, can simplify the Precondition check

PR #543, opened 2026-03-08
Fix: Linux AppImage nesting causes slow file picker in dolphin

ZadenRB:
Launching the mainline beta from the Slippi launcher caused dolphin file pickers to be extremely slow to open & navigate. Root cause was the AppImage environment of dolphin mainline also including environment variables from the Slippi launcher AppImage. This led to various icons being searched for in the Slippi AppImage directory rather than the proper user directories:
```
# There are thousands of these syscalls upon opening a file picker in dolphin mainline
# /tmp/.mount_SlippiJ7w6Ag is the mount point of the launcher AppImage

access("/tmp/.mount_SlippiJ7w6Ag/usr/share/icons/hicolor/96x96@2/actions/text-plain.png", F_OK) = -1 ENOENT (No such file or directory)
```

Originally encountered on Fedora 43 KDE Plasma 6.6, Reproduced on Debian 14 KDE.

This PR resolves the issue by spawning the dolphin process with a cleaned environment, where any references to the parent app image mount point removed.

### DIFF
--- a/src/dolphin/instance.ts
+++ b/src/dolphin/instance.ts
@@ -102,11 +102,14 @@ export class DolphinInstance extends EventEmitter {
           });
         } else {
           const appdir = process.env.APPDIR;
+          const dolphinEnv = {};
           for (const key in process.env) {
             const split_val = process.env[key].split(":");
-            process.env[key] = split_val.filter((val) => !val.includes(appdir)).join(":");
+            dolphinEnv[key] = split_val.filter((val) => !val.includes(appdir)).join(":");
           }
-          child = spawn(executablePath, params);
+          child = spawn(executablePath, params, {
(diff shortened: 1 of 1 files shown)


NikhilNarayana:
i'm not entirely against this, but i kinda wanted to fix mainline not opening the default file picker. if i can't figure it out there by the end of the weekend then i'll merge this for now

ZadenRB:
> i'm not entirely against this, but i kinda wanted to fix mainline not opening the default file picker. if i can't figure it out there by the end of the weekend then i'll merge this for now

Fair enough - I'm also going to look into the styling some more this weekend since I have some free time.

vinceau:
Superseded by #608

PR #570, opened 2026-05-08
feat: Display user's current rank on the home page

vinceau:
### Description

This PR adds the current user's rank and rating to be displayed in the launcher. The logic is mostly copied from the slippi website, and the design is again [inspired from the Figma](https://www.figma.com/design/GHmNu63U9zaZi6jD7zK8ow/Slippi-Launcher?node-id=137-204&t=0cDqypXK8iQ7piJg-0).

We add a button to allow the user to refresh the data. There is no auto-refresh enabled (I don't think we need this?), if we do what's a good refresh rate? I think it's fine to have the user refresh as desired.

### Screenshots
Screenshot 2026-05-08 at 5 46 40 pm

Screenshot 2026-05-08 at 5 47 04 pm

Screenshot 2026-05-08 at 5 46 19 pm
Screenshot 2026-05-08 at 5 47 50 pm

Screenshot 2026-05-08 at 5 47 27 pm


JLaferri:
Looking good overall! I have some minor styling concerns, curious if you agree or not.

1. I think the former Ranked Day card looked a bit more cohesive. In the new one I think the green left aligned header kind of conflicts with the purple and center aligned rest of the card. I feel like either we just remove the header from that card and style it like before (think this is my preference?). Or maybe we try left aligning everything else in the card?
2. The white space above and below the ranked day display at large sizes currently disconnects if from the fact that it's ranked day. I think it'd look cleaner if we just made the whole card smaller and left the whitespace in the background instead of making the card the bigger?
3. I like how the ranking display looks at small sizes but it feels a bit less cohesive at large sizes where the ranked icon is so much bigger than the ranked text. Maybe we could just limit the max height more so it doesn't get so large?

vinceau:
1. I agree with you there. I was mostly trying to stick close to the Figma designs where possible which is why I've had them left aligned. I can try out a few different layouts like the ideas you've mentioned and see what you think.
2. I can limit the size of the cards and bring back the whitespace.
3. I'm happy to keep this smaller too.

vinceau:
How's this @JLaferri ?


Screenshot 2026-05-13 at 6 27 03 pm
Screenshot 2026-05-13 at 6 27 08 pm

JLaferri:
> How's this @JLaferri ?
> 
> Screenshot 2026-05-13 at 6 27 03 pm Screenshot 2026-05-13 at 6 27 08 pm

Much better imo, thanks! I think the rank max height could be even smaller but I'm starting to nitpick, it already looks quite solid imo.

NikhilNarayana (approved):
happy with it, one possible idea for auto refreshing the rank is to do it on netplay dolphin close, but that's a minor suggestion

PR #575, opened 2026-05-12
fix: Dolphin installation blocking main thread on MacOS

vinceau:
## Description

This PR fixes the issue where Dolphin installation blocks the main thread - causing the renderer to stop responding.

We achieve this by:
* using spawn instead of `spawnSync` to avoid blocking main
* making use of more concurrent asynchronous operations
* yielding to the event loop more often

We test this by:
* Clearing the downloaded Dolphins and temp folder from `~/Library/Application Support/Slippi Launcher`
* Opening the app and repeated scrolling

I've only tested this on MacOS. On my M5 MBP it would originally stall for approximately 11 seconds. On the fixed version it does not stall at all.

@NikhilNarayana tested Windows and Linux and confirmed that this problem is localised to only MacOS.

### Current behaviour (Blocks main thread)

https://github.com/user-attachments/assets/3070ca35-aa08-46d5-8237-eb52cf156e06

### New behaviour (main thread is unblocked)

https://github.com/user-attachments/assets/541d9bff-b340-407f-a33b-a27a4197ee84

### DIFF
--- a/src/dolphin/install/execute_command.ts
+++ b/src/dolphin/install/execute_command.ts
@@ -1,0 +1,38 @@
+import { spawn } from "child_process";
+
+/**
+ * Execute a command and return its output
+ * @param command The command to execute
+ * @param args The arguments to pass to the command
+ * @returns The output of the command
+ */
+export async function executeCommand(command: string, args: string[]): Promise<string> {
+  return new Promise((resolve, reject) => {
+    const child = spawn(command, args);
+    let stdout = "";
+    let stderr = "";
+
+    child.stdout?.on("data", (data) => {
+      stdout += data.toString();
+    });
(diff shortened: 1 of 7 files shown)


NikhilNarayana (commented):
i'm on the fence about all the extra logging. i'm not sure i would see much value from it from a support perspective. The `Extracting:` logs need to be removed though, they're far too noisy

vinceau:
Yeah I'm happy to remove the logging.

NikhilNarayana:
cool then let's just do that. the only other possible suggestion i have is putting _executeCommand somewhere common since it's identical in both places

NikhilNarayana (approved):

PR #576, opened 2026-05-12
feat: Add dual pane news feed and update news preview styling

vinceau:
## Description

This PR adds a dual pane news feed to the Latest News tab, and updates the news preview in the overview. This combined with #570 completes our home page revamp efforts.

We also introduce read/unread status. We define "new" news as those posted within the last month and mark those as unread by default. This prevents the issue where a giant news list gets hard to manage and remember what has or hasn't been read.

I chose not to add a "Mark all as read" button because I don't think there will ever be more than a small handful of unread news at a given time. If 4-5 items is too many to show at once (only really matters on first release) I can reduce the window from 1 month to 2 weeks. This should reduce the number of unreads down to maybe 1-2 items.

We also add icons to easily identify which posts come from what news source.

## Screenshots
Screenshot 2026-05-12 at 5 59 42 pm
Screenshot 2026-05-12 at 5 59 51 pm
Screenshot 2026-05-12 at 6 00 50 pm
Screenshot 2026-05-12 at 6 00 56 pm
Screenshot 2026-05-12 at 6 01 39 pm


NikhilNarayana (approved):
i'm in favor of making the newest post the default selection rather than nothing, but i kinda understand why didn't do that. the look and feel otherwise is great

PR #584, opened 2026-05-21
i18n: update Portuguese translations

vinceau:
This PR adds Portuguese machine translations for the missing strings. Will await @CaioIcy to double check these before merging.


CaioIcy (approved):
LGTM, thanks

PR #586, opened 2026-05-22
fix(security): 2 improvements across 2 files

tomaioo:
## Summary

fix(security): 2 improvements across 2 files

## Problem

**Severity**: `Medium` | **File**: `src/dolphin/install/execute_command.ts:L14`

The `executeCommand` function in `src/dolphin/install/execute_command.ts` uses `spawn` with a command string and args array. While `spawn` with array arguments is generally safer than `exec`, the function is exported and could be called with untrusted input. More critically, the function resolves on `code === 0 || output` which means commands that fail with non-zero exit codes but produce any output (even to stderr) are treated as successful. This could mask failures or allow unexpected behavior if callers rely on exit codes.

## Solution

Add input validation for the `command` parameter to ensure it only contains expected commands. Consider using a whitelist of allowed commands. Also, reconsider the success condition - failing commands should probably not be treated as successful just because they produced stderr output.

## Changes

- `src/dolphin/install/execute_command.ts` (modified)
- `.erb/scripts/windows-sign.js` (modified)

### DIFF
--- a/.erb/scripts/windows-sign.js
+++ b/.erb/scripts/windows-sign.js
@@ -1,7 +1,7 @@
 /* eslint-disable @typescript-eslint/no-var-requires */
 // Custom Sign hook for use with electron-builder + SSL CodeSignTool
 
-const { execSync } = require("child_process");
+const { execFileSync } = require("child_process");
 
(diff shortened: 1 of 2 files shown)


NikhilNarayana:
no

PR #589, opened 2026-05-23
Add hide rank display toggle

zaderrr:
Adds a toggle to the Appearance settings of the launcher to hide the users current rank display.
Added a new setting `enableRankDisplayCard` to the settings schema, defaulted to True.

This only affects the launcher, leaving the existing in-game setting unchanged.

In regards to translations, I ran both `i18n:extract` and `i18n:sync` which looks to have formatted existing translations onto new lines. As far as I can tell, it's displaying "correctly" in alternative languages, though I don't speak any of these languages. I'm not really familiar with localization on projects, so unsure on how this process works.

image

Toggled off:
image


vinceau:
Thanks for looking into this.

Regarding i18n, you don't have to worry about running the i18n commands for general feature additions. We'll handle the extraction and sync. You can update the messages and that's it.

A couple thoughts: 
* Can you look into fetching and displaying the value in their dolphin settings? 
* I'm currently thinking of having maybe a dropdown for a setting "Hide ranking" with maybe a dropdown with 3 options "In launcher", "In dolphin", "always", "never". but I'm not sure if that's the best way to display that information. Will discuss it with the team. Or alternatively we could have a "Show ranking" with two checkboxes, one for the launcher and one for dolphin, with both defaulted to on.

src/renderer/pages/settings/appearance_settings/appearance_settings.tsx
```diff
@@ -1,11 +1,13 @@
 import React from "react";
 
 import { LanguageSelector } from "./language_selector/language_selector";
+import { RankDisplayToggle } from "./rank_display_toggle/RankDisplayToggle";
```
vinceau:
```suggestion
import { RankDisplayToggle } from "./rank_display_toggle/rank_display_toggle";
```
All files should follow snake casing.

src/renderer/pages/settings/appearance_settings/rank_display_toggle/rank_display_toggle.messages.ts
```diff
@@ -0,0 +1,4 @@
+export const RankDisplayToggleMessages = {
+  rankDisplay: () => "Rank display",
+  rankDisplayDescription: () => "Enable the current rank display on the home page.",
```
vinceau:
```suggestion
  rankDisplay: () => "Show Ranking on Home Page",
  rankDisplayDescription: () => "If enabled, your current rank and rating will be displayed on the Home page.",
```

src/settings/types.ts
```diff
@@ -65,6 +65,9 @@ export interface SettingsSchema {
   // Spectate settings
   enableSpectateRemoteControl: boolean;
   spectateRemoteControlPort: number;
+
+  // Appearence settings
+  enableRankDisplayCard: boolean;
```
vinceau:
```suggestion
  enableRankDisplay: boolean;
```
I would drop all references to "Card". Card is a frontend term for a UI component, I don't think it should be tied to the setting name.

src/renderer/pages/settings/appearance_settings/rank_display_toggle/rank_display_toggle.tsx
vinceau, line 1: this file should be called `rank_display_toggle.tsx`

src/renderer/lib/hooks/use_settings.ts
```diff
@@ -95,6 +95,8 @@ export const useAutoUpdateLauncher = () => useSetting("autoUpdateLauncher");
 
 export const useEnableLocationAccess = () => useSetting("enableLocationAccess");
 
+export const useEnableRankDisplayCard = () => useSetting("enableRankDisplayCard");
```
vinceau:
```suggestion
export const useEnableRankDisplay = () => useSetting("enableRankDisplay");
```
Again I would drop all references to "Card"

vinceau (changes requested):
Looks good for the most part with some minor changes. Can you also revert the entire `locales` folder? (like via `git checkout main -- locales`).

To summarise discussion that happened on Discord, we'll just leave the Dolphin setting in Dolphin, and keep the settings separate and unrelated. As a result, we'll also keep the toggle under the Appearance section.

zaderrr:
Requested changes made, thanks for the review.

vinceau (approved):
LGTM. Thanks for your contribution!

Issue #180, opened 2021-06-23
Refactor: Stats Page


NikhilNarayana:
Adding L-Cancels to the stats page was not easy and showed that the stats page was in need of a refactor. Code design should be discussed with Vince and I.

CMarah:
@NikhilNarayana Is it ok if I open a PR for this issue and we discuss there?

vinceau:
Completed in #600

PR #600, opened 2026-06-02
refactor: Improve stats page readability and maintainability

vinceau:
Addresses #180


src/renderer/pages/replays/replay_file_stats/overall_table/build_sections.ts
```diff
@@ -0,0 +1,159 @@
...
+  };
+}
+
+export function buildSections(p1: PlayerStatSummary, p2: PlayerStatSummary): StatSection[] {
+  return [
+    {
+      title: "Offense",
+      rows: [
+        simpleStat("Kills", p1.kills, p2.kills, (a, b) => a > b, String),
+        simpleStat("Damage Done", p1.totalDamage, p2.totalDamage, (a, b) => a > 0 && b > 0 && a > b),
```
zaderrr: If a or b == 0, this causes no highlighting even if a player has  > 0
Image

(a > 0 || b > 0) && a > b, perhaps?
Image
Image

src/renderer/pages/replays/replay_file_stats/overall_table/build_sections.ts
```diff
@@ -0,0 +1,159 @@
...
+        fractionStat("Opening Conversion Rate", p1.successfulConversions, p2.successfulConversions, "higher"),
+        simpleStat("Openings / Kill", p1.openingsPerKill, p2.openingsPerKill, (a, b) => a < b),
```
zaderrr: If a or b == 0, and the other > 0, then highlight picks up the lower value still (being 0), causing incorrect highlight (I believe), with 0 being defaulted when no stocks taken/undefined Image

zaderrr (commented):
Hi, two small logic issues with highlighting I found.

src/renderer/pages/replays/replay_file_stats/overall_table/build_sections.ts
vinceau: should be fixed now. i don't think it needs the zero check at all.

src/renderer/pages/replays/replay_file_stats/overall_table/build_sections.ts
vinceau: should be fixed now, just ensured that the value itself is non-zero.

PR #603, opened 2026-06-03
feat: console auto connect

jhudson1998:
Adds an opt-in Auto-connect toggle to the Console Mirror page. When enabled, saved consoles are connected automatically as they appear on the local network, so you no longer have to click Connect on each one. Consoles disconnected manually stay disconnected during that session. (Unless Auto-connect is toggled off/on again in which case it will connect to all disconnected consoles)

**Use case**

We run a fully networked setup at our local melee events with 6-12 Wiis connected via USB-LAN adapters for gathering replays + streaming. Our local scene has upwards of 20 different Wiis that are used at any given time meaning that the list shown in the launcher has grown quite extensive and provides some friction when connecting. This would reduce the time it takes to connect each Wii individually and just allow us to just open the launcher and have the saved Wiis on the network connect automatically.

**Behavior**

- Toggle ON: automatically connects every saved console that is currently broadcasting, not already connected, and not manually disconnected this session
- Toggle OFF: stops auto-connecting, existing connections untouched
- Manual Disconnect: opts that console out of auto-reconnect for the rest of the session (or until Auto-connect is toggled off/on)
- Manual Connect: opts the console back in
- Re-enabling the toggle: clears manual opt-outs and re-connects all saved consoles as if starting fresh
- A console that drops off the network on its own (compared to a manual disconnect) is not opted out, so it reconnects when it reappears
- Toggle defaults to off, existing UX is retained until a user enables the toggle
- Opt-out lifetime: The manual-disconnect opt-out lives in the discovery store & survives page navigation. Cleared on app restart and when the toggle is re-enabled

**TBD**
- Not sure about the styling/placement on the toggle, reused toggle styling + placed at top of console list but looking for input on the best way to display
- I think a filter for Available consoles would be a great addition to this feature but I'm unsure how it would fit into the current UI, open to thoughts on that
- Translations TBD as the text I added for the toggle should be confirmed first

https://github.com/user-attachments/assets/969d2e39-ec27-451f-a47d-c05ed67bd906


src/renderer/pages/console_mirror/saved_connections_list/saved_connection_item.tsx
```diff
@@ -51,35 +52,21 @@ export const SavedConnectionItem = ({
...
+    allowAutoConnect(connection.ipAddress);
+    await consoleService.connectToConsoleMirror(buildMirrorConfig(connection, nickname));
+  }, [consoleService, connection, nickname, allowAutoConnect]);
   const onMirror = React.useCallback(() => {
     consoleService.startMirroring(connection.ipAddress).catch(showError);
   }, [consoleService, connection, showError]);
-  const onDisconnect = () => consoleService.disconnectFromConsole(connection.ipAddress);
+  const onDisconnect = () => {
+    // Manually disconnecting opts the console out of auto-connect for this session.
+    optOutOfAutoConnect(connection.ipAddress);
```
vinceau, line 67:
```suggestion
    optOutOfConsoleAutoConnect(connection.ipAddress);
```

src/renderer/pages/console_mirror/console_mirror.tsx
```diff
@@ -14,10 +14,17 @@ import { DualPane } from "@/components/dual_pane";
...
-import { addConsoleConnection, deleteConsoleConnection, editConsoleConnection } from "@/lib/console_connection";
+import {
+  addConsoleConnection,
+  buildMirrorConfig,
+  deleteConsoleConnection,
+  editConsoleConnection,
+} from "@/lib/console_connection";
 import { useConsoleDiscoveryStore } from "@/lib/hooks/use_console_discovery";
-import { useSettings } from "@/lib/hooks/use_settings";
+import { useEnableAutoConnect, useSettings } from "@/lib/hooks/use_settings";
```
vinceau, line 27:
```suggestion
import { useEnableConsoleAutoConnect, useSettings } from "@/lib/hooks/use_settings";
```

src/renderer/lib/auto_connect.ts
vinceau, line 1: Should be called `console_auto_connect.ts`

vinceau (changes requested):
Can you rename everything that is just "auto connect" to be "console auto connect"? Just "auto connect" is too ambiguous. I've added a few correction examples but it is not exhaustive.

Also I've rebased your branch off main, and you should be able to see that there is a toggle now on the spectate page. Can you style your toggle to be based on that in the spectate page? For the long description, you can have the helper button like in the "What is Mirroring" `<MirroringHelp />` button.

PR #608, opened 2026-06-05
Sanitize AppImage environment when launching Dolphin

ellierocks:
When Slippi Launcher starts as an AppImage, it sets environment variables like APPDIR, APPIMAGE, LD_LIBRARY_PATH, and QT_PLUGIN_PATH. Those can leak into Dolphin AppImages and cause Dolphin to resolve libraries or Qt plugins from the launcher paths instead of its own. This results in many distros being unable to select mainline due to the QT_PLUGIN_PATH being read as "".

This applies the sanitized environment to both Dolphin launch and Dolphin version probing. Another fix is also required on dolphin mainline to fully solve this issue.


vinceau:
Is this related or the same issue as #543?

ellierocks:
@vinceau most likely yes, I didn't see that one but it seems these are caused by the same issue. I came up with this fix from my testing on arch, since the appimage there currently can't use mainline at all without messing with environment variables.

vinceau (approved):
Added some tests and renamed the function and the file. Thanks for the contribution!

